PR #8, opened 2019-10-28
update pulse sms

Armaldio:
Just updating the URL of the service


kytwb:
Thanks for your contribution, it's more clear using this URL 👍

PR #11, opened 2019-11-27
Override user agent in Gmail recipe

kytwb:
Fixes https://github.com/getferdi/ferdi/issues/140 and potentially https://github.com/getferdi/ferdi/issues/229.
As suggested in https://github.com/ramboxapp/community-edition/issues/2385#issuecomment-526737000.

Before:

67165100-0302ec80-f34f-11e9-8b80-1d56dca3df2f

After:

69717909-9219d600-113f-11ea-97c4-439c64ac1d6c


kytwb:
@vantezzen can we deploy this to current users without https://github.com/getferdi/server/issues/14?

vantezzen:
No, I think service updates need to be implemented first. We could however push it into the next Ferdi Release so people who create the service from scratch or those who delete the recipe folder will already have the new version

kytwb:
Understood, let's do like this then so we can deliver the fix for now. Can I merge the open pull requests?

PR #39, opened 2020-02-11
Update slack: Fix incorrect notification count

Serubin:
This PR includes https://github.com/meetfranz/recipe-slack/pull/13

This also makes some minor other minor updates to the slack plugin that were included in upstream.

## The Problem
Currently slack will show 1 notification even if there are zero. This is due to a code change in Slack.

## The Solution
I changed the current CSS selector ignore the "All Unreads" channel.


Serubin:
Fixed the crazy formatting changes - please squash merge :)

vantezzen:
Could you please increase the version number for the recipe and package again? Otherwise Ferdi won't update the recipe.

Serubin:
You got it @vantezzen. Thanks for catching that!

vantezzen:
Looks like you didn't update the version number correctly yet - Ferdi will fall into an update loop like this.
Please load the recipe into `scripts/recipe_src` (do `yarn load slack` if its not already loaded), update the version number in `scripts/recipes_src/package.json` and `yarn package` again to make sure the version number inside the recipe itself is the new version number as well.

Serubin:
Apologies, should be good now.

vantezzen:
Sorry, still not properly updated: `/slack.tar.gz` still contains the old version number, thats why you have to repackage and not edit the unpackaged file.
I just updated the version number and will merge from https://github.com/getferdi/recipes/tree/pr/39

PR #44, opened 2020-02-18
[discord] Fix indirect notifications, simplify code

dpeukert:
Discord changed their unread favicon, I've kept the old one in there in case it's a partial rollout.


vantezzen:
Don't know why GitHub is executing that action on that point but your PR looks good. Thank you for your contribution.

PR #46, opened 2020-02-21
Feat/gmeet

n-xlkt:


vantezzen:
Thank you for your contribution.
Could you please replace the `index.js` with https://github.com/getferdi/recipes/blob/master/uncompressed/gmail/index.js ? We had some problems with Google when not overwriting the user agent to that.
Also, you might want to change `webview.js` so it uses [indirect messages](https://github.com/meetfranz/plugins/blob/master/docs/frontend_api.md#setbadgedirectmessages-indirectmessages) for muted chat instead of adding them to the direct messages, what do you think?

Other than that, it looks great.

n-xlkt:
> Also, you might want to change `webview.js` so it uses [indirect messages](https://github.com/meetfranz/plugins/blob/master/docs/frontend_api.md#setbadgedirectmessages-indirectmessages) for muted chat instead of adding them to the direct messages, what do you think?

as _google meet_ only notifies you of meetings you've decided not to decline from your calendar, i can't think of any instances of indirect messages.

vantezzen:
> as google meet only notifies you of meetings you've decided not to decline from your calendar, i can't think of any instances of indirect messages.

Oh sorry, I misunderstood that feature there.

It looks like you havn't repackaged your changes again. Please copy the recipe into `scripts/recipe_src` again and run `yarn package`. Ferdi only uses the recipes in the `.tar.gz` archives, the uncompressed recipes are only for better access to the code from the repo and have no use otherwise.

PR #51, opened 2020-02-28
Add GitHub Entreprise recipe

antoine-aumjaud:
Add GitHub Entreprise recipe


kytwb:
Thanks for your contribution @antoine-aumjaud! 🙏

PR #65, opened 2020-03-11
Add Unraid + Netatmo Energy recipes

rctneil:
My first attempt at a plugin. Not sure if i've done it correctly but it's my first attempt!


uncompressed/unraid/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "id": "unraid",
+  "name": "Unraid",
+  "version": "1.0.1",
+  "description": "Unraid dashboard",
+  "main": "index.js",
+  "author": "Neil Tonge <me@neiltonge.co.uk>",
+  "license": "MIT",
+  "repository": "https://github.com/rctneil/franz-recipe-unraid",
+  "config": {
+    "serviceURL": "http://{teamId}.local",
```
vantezzen: The unRAID doesn't always have to be a `.local` domain, does it? It could be an IP or other domain instead.

I suggest changing the serviceURL to just `{teamId}` so that users can input any URL or IP they want.

rctneil: Updated! Do i need to submit a new pull request?

kytwb: @rctneil No, you can continue to push edits on the same branch, it will add the edits you push to the pull request 🙂

rctneil: Ok, Just left it as is. I take it there's nothing else i need to do then? Hopefully it'll get accepted!!!

kytwb: @rctneil it looks like you still need to take @vantezzen feedback above in consideration 😄 (regarding the `serviceURL`).

rctneil: Oh i'm sorry, I did change that but maybe I forgot to push the change.  I will push it when I get home.  Apologies!

kytwb: @rctneil No worries! 🙏 We appreciate your contribution!

vantezzen:
Checks are failing - this means that your uncompressed folder at `/uncompressed/<service ID>` does not have the same content as your archived recipe. Please make sure both are in sync (it looks like you maybe didn't update the uncompressed version correctly.)

rctneil:
@vantezzen Apologies.  I had made changes but they hadn't been pulled in.   I THINK it's correct now but if not just let me know and i'll give it another shot!   If it's correct, how do I see the new recipes become available in Ferdi?

rctneil:
> Looks good to me. Have you tested these recipes locally if they really work?

Yes, I actually wrote them a while back for Franz and used them for a short while but then Franz started to act up, on my machine by hogging CPU etc. I gave up on Franz as I saw it wasn't getting updated and tried to find an alternative, I failed. Someone then pointed me to Ferdi and i'm back to being happy.   I thought yesterday why not try to get them officially added and here we are!

I have noticed that the notifications for Unraid can be a little dodgy which is why i've had to grab the number of notifications and minus 1 but I will keep an eye on it and if the issue at their end gets resolved I will update my recipe.

Netatmo Energy doesn't seem to do any form of notifying so it's a very basic web view but as it's a dashboard it's nice to have it in Ferdi.

How do I see the new recipes become available in Ferdi?

vantezzen:
Then welcome to the Ferdi community!

Your recipes will be shipped with the next update of Ferdi (we actually just released a new beta version a few days ago so it will be some time until the next one gets released).

rctneil:
@vantezzen Fantastic!   Thankyou. It's great to see Ferdi in active development.  One thing i've noticed is that sometimes it can act rather unresponsive or slow at times.  Also happens when attempting to open the Settings window.

vantezzen:
@rctneil This usually happens when you have a lot of services open or when you don't have many CPU or RAM resources available.

Have you tried enabling service hibernation so services you don't currently use need less resources? Also, you can use the process manager (open settings and scroll down to the button "Open Process Manager") to see what process of Ferdi is using the most resources and may be causing your issues.

PR #69, opened 2020-03-20
Make the recognition of the GMail's landing page stricter

tpopela:
Extend the check to include the location's href. Checking only for
hostname could activate the relocation even in cases when it's not
desired - such as after authenticating with SAML and relocating back
to GMail.

Closes: getferdi/ferdi#473


vantezzen:
Closing this but merging #70 in order to correctly package the recipe

PR #80, opened 2020-03-26
Fix icon for Slack March 2020 styles update

justinebateman:
Slack is rolling out a major styles update. This is breaking the selector for the icon in the Ferdi sidebar. I've updated the selector to cover both the old Slack styles and the new ones because not all workspaces have been migrated yet


vantezzen:
Thank you for your contribution! I'll close this PR but will merge your changes via e739f16 to correctly update and repackage the recipe files.

justinebateman:
Thanks a lot! I'll keep that in mind for next time

PR #89, opened 2020-04-01
Don't set Mac UA on Linux for Google services

tpopela:
If we set the Mac UA on Linux, then various shortcuts will change to Mac
ones - i.e. Ctrl + K to create a link in GMail will stop work, Ctrl
+ Enter to send an email in GMail and others. Only set the Mac UA when
  not running on Linux.


vantezzen:
Thank you for your contribution. Unfortunately, it looks like you havn't correctly packaged and updated the recipe files:
Please repackage the recipes instead of just editing the uncompressed versions - Ferdi doesn't use uncompressed recipes!
To repackage, please load a recipe using e.g `yarn load gmail`, update your code in `scripts/recipe_src` and **increase the version number** in `package.json` - Ferdi uses this version number to check if it should update the recipe - and repackage the recipe using `yarn package`.

tpopela:
> Please repackage the recipes instead of just editing the uncompressed versions - Ferdi doesn't use uncompressed recipes!
> To repackage, please load a recipe using e.g `yarn load gmail`, update your code in `scripts/recipe_src` and **increase the version number** in `package.json` - Ferdi uses this version number to check if it should update the recipe - and repackage the recipe using `yarn package`.

Ok, this was the information I was missing from the documentation (https://github.com/getferdi/recipes/blob/master/docs/integration.md). It didn't said anything about fixing the current ones, but talked about uploading a new recipes.

vantezzen:
Sorry, we'll improve the documentation on that in the future.
Thank you again for your contribution!

PR #94, opened 2020-04-10
Add Nextdoor recipe

JakeSteam:
Adds support for [Nextdoor](https://en.wikipedia.org/wiki/Nextdoor).

Since it's not available in all regions, and it's hard to test, here's a screenshot of it in action:
image

Made on Windows, you'll be pleased to hear all the publishing steps seem to work fine (after installing yarn)!


vantezzen:
Thanks for the contribution! Glad to hear that packaging works on Windows.

PR #103, opened 2020-04-15
Add new Zalo recipe

RoiArthurB:
Zalo is a very popular Vietnamese chatting app with a web version :)
https://en.wikipedia.org/wiki/VNG_Corporation#Zalo


vantezzen:
Thank you for your contribution. This recipe looks mostly good, but your `icon.svg` and `icon.png` are currently 2500x2372 pixels but Ferdi requires them to be 1024x1024 pixels (otherwise the icon will not be displayed correctly).
Could you please resize both files using something like Photoshop, GIMP or [Photopea](https://www.photopea.com/)? Please also keep in mind that you'll have to _repackage_ the recipe after changing the icons, again using `yarn package`.

RoiArthurB:
> Could you please resize both files using something like Photoshop, GIMP or Photopea?

Ok I'll do

> Please also keep in mind that you'll have to repackage the recipe after changing the icons, again using `yarn package`.

Means I should this command instead of `yarn github ...` ?

vantezzen:
Sorry, if you are using `yarn github ...` then please use that again, `yarn package` is only for when you are not importing recipes from GitHub.

RoiArthurB:
Done ! :)

vantezzen:
Great. Again thank you for your contribution, your recipe will be available in the next version of Ferdi.

PR #116, opened 2020-04-19
Update Microsoft Teams to allow Desktop Sharing

gautamsi:
Should fix part of https://github.com/getferdi/ferdi/issues/522 (Screen Sharing)

I tried this recently and could not share desktop. seems like you have to have recent Chrome version and enable support for screen sharing selection, Chrome and Edge does that with their own UI. 

came across a solution which is focused for polyfill for Chrome specific api in Electron. 

* Update Teams user agent (no longer need to set Chrome version to 63.x)
* Enable Desktop sharing per https://github.com/electron/electron/issues/16513#issuecomment-602070250

I am new to this contribution here. let me know if there is better way to write some of the helper things. It would be better to see per service (teams instance) config to do this. but for now it works that way.


all.json
```diff
@@ -510,7 +510,7 @@
     "featured": false,
     "id": "msteams",
     "name": "Microsoft Teams",
-    "version": "1.2.0",
+    "version": "1.3.0",
```
gautamsi, line 513: does this require a major version bump?

uncompressed/msteams/index.js
```diff
@@ -2,7 +2,6 @@
 
 module.exports = Franz => class MicrosoftTeams extends Franz {
   overrideUserAgent() {
-    return window.navigator.userAgent.replace(/(Ferdi|Electron)\/\S+ \([^)]+\)/g, '').replace(/(Chrome\/)([^ ]*)/g, '$163.0.3239.84');
+    return window.navigator.userAgent.replace(/(Ferdi|Electron)\/\S+ \([^)]+\)/g, '');
```
gautamsi, line 5: for me it did not work properly for having chrome 63 version. So I removed it. I have no context on why do we need this to specific Chrome version, probably earlier version of Electron was on lower chrome version and it needed to boost the version for Teams to recognise supported chrome version.

uncompressed/msteams/service.css
```diff
@@ -1,3 +1,63 @@
 notification-banner {
 	display: none !important;
 }
+.desktop-capturer-selection {
```
gautamsi, line 4: additional classes to work with desktopsharing picker

uncompressed/msteams/webview.js
```diff
@@ -1,7 +1,64 @@
 "use strict";
 
+const { desktopCapturer } = require('electron');
 const path = require('path');
 
+window.navigator.mediaDevices.getDisplayMedia = () => {
```
gautamsi, line 6: polyfill per the Electron issue comment

kytwb:
@gautamsi Thank you so much for taking a look at this! 💖 I am not sure if it is best implemented at the recipe level, or do you think it should be implemented within Ferdi directly? Right now we don't have a way to know which service requires screen sharing, but if we had, couldn't we implement once in Ferdi and load/inject what's needed for the screen sharing dialog? 

@vantezzen do you think we should take the pull request as it and replicate where needed in other recipes for now?

all.json
kytwb: @gautamsi We don't properly take semver in consideration right now, we just need to bump the version to trigger a service update/reload in the clients. Minor is fine here in my opinion 🙏

uncompressed/msteams/index.js
kytwb: Thank you for sharing your thought process 🙏

gautamsi:
If you do it with Ferdi code, then each service would need to have a toggle button to enable screen sharing for security. Some more control may be added via `webContents.session.setPermissionRequestHandler` handlers in `MainWindow` instance which was needed mentioned in https://github.com/electron/electron/issues/16513#issuecomment-602070250 (I did not have to use that piece of code to enable it in Ferdi.)

I see that it may also benefit Skype, Slack, Hangout, Google Meet and possibly other non-Electron aware messengers which allow screen sharing. I had need to use multiple Teams tenant (not a in built feature to Teams yet - they are working on it) so I came looking for this, troubleshooted full day on this and figured out a way to allow screen sharing on this. I have not tested other apps.

kytwb:
@gautamsi Thank you for the detailed answer! Let's ship then so other Teams users can already start benefit from it; we can use as reference in another iteration to port to other services.

PR #124, opened 2020-04-20
Add support for indirect messages in MS Teams

FeikoJoosten:


kytwb:
Thank you! 🙏 Apologies for the broken builds, will investigate via https://github.com/getferdi/recipes/issues/130.

FeikoJoosten:
Np!

PR #111, opened 2020-04-17
Update Instagram recipe to use official web service

kytwb:
This uses https://github.com/meetfranz/recipe-instagram
For https://github.com/getferdi/recipes/issues/59

PR #126, opened 2020-04-20
Remove misplaced Instagram logos in root.

halms:
The last commit updating the instagram recipe somehow added instagram logos in the root.
Let's remove them again.


vantezzen:
Thank you for this contribution. This must have been introduced with #111.

PR #127, opened 2020-04-20
Update Slack: notifications to exclude draft messages

Serubin:
Updated CSS selector for notifications to exclude "drafts" in new slack UI. 

This should be backward compatible with the older version of slack.


vantezzen:
Thank you for your contribution. This change will be shipped with the next version of Ferdi.

PR #129, opened 2020-04-21
Adds notifications for GitHub

FeikoJoosten:
I've added notifications support for github. Currently it simply checks if the bell icon in the top right has unread messages. It doesn't properly check the amount of notifications. 
image
image


kytwb:
@FeikoJoosten wouldn't it make sense to leverage indirect notifications if we don't count the amount of notifications then?

kytwb:
Indirect notifications badge color is changing to blue with the next release too (via https://github.com/getferdi/ferdi/pull/590).

FeikoJoosten:
Hmm yeah making use of the indirect notifications might make more sense in this case. Easy fix ;) 

And the indirect notification color changing is something I’m aware off. I suggested it (and helped implementing it 😜)

I’ll change it later so it uses indirect notifications instead. It’s basically a matter of adding `0, ` in the `setBadge` function.

FeikoJoosten:
@kytwb and fixed ;)

kytwb:
@FeikoJoosten thank you! 

Can I ask how do you package the recipes? Are you using `yarn run package` as described in https://github.com/getferdi/recipes/blob/master/docs/integration.md#publishing? I had to repackage them manually via https://github.com/getferdi/recipes/pull/131.

FeikoJoosten:
Hmm that's odd. I'm using `npm run package`. So should be the same thing? 🤔

kytwb:
Let's continue in https://github.com/getferdi/recipes/issues/130, I will merge the changes so we can ship 5.5.0-gm.2 with all the fixes/improvements 🙏

uncompressed/github/webview.js
```diff
@@ -5,6 +5,6 @@ module.exports = Franz => {
     Franz.setBadge(document.querySelector('[class*="mail-status unread"]') !== null ? 1 : 0);
   };
 
-  Franz.loop(getMessages);
+  Franz.loop(0, getMessages);
```
kytwb, line 8: @FeikoJoosten I just noticed; I think you wanted to add that `0, ` to setBadge right?

FeikoJoosten: Yep, wanted to do that... Not sure how I managed to add it to the loop instead. Must've been a bit too excited haha. I've fixed it with another commit. But based on what you've done in #133 I don't think it was necessary. Thanks for the review though! I'll review the PR you requested my review on in a bit 😄

Issue #130, opened 2020-04-21
Error while reviewing recipes


kytwb:
**Describe the bug**
Occurred in on-going pull request: https://travis-ci.org/github/getferdi/recipes/builds/677417795#L229

**Expected behavior**
I double checked the archive it looked good to me, should have worked.

kytwb:
Looks like it similar to https://github.com/getferdi/recipes/pull/98#issuecomment-613404716.

kytwb:
Cancelled by https://github.com/getferdi/recipes/pull/292.

PR #131, opened 2020-04-21
Repackage archives not passing the checks

kytwb:
For https://github.com/getferdi/recipes/issues/130. I had to manually repackage the recipes not passing the checks. Not sure why.


uncompressed/discord/webview.js
```diff
@@ -5,11 +5,6 @@ var _path = _interopRequireDefault(require("path"));
 function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
 
 module.exports = Franz => {
-  const iconValues = [
```
kytwb, line 8: @vantezzen I also removed those as it was unused and it got me stopping when manually double checking.

Issue #128, opened 2020-04-21
Switch back to the original Telegram from Telegram React?


lolmaus:
I don't know, why the switch to Telegram React was made. I have not found a single thing that is better in Telegram React.

But it has a few issues, I've ran into two:
— On a window 1050px wide the list of chats collapses and chat names are not visible. 
— It is impossible to mention a user because autocomplete does not work.

How do I switch back to web.telegram.org? Why is there no choice between the two?

vantezzen:
Related: #95. I'd suggest moving the current recipe to "telegram-react" and restoring the telegram recipe.

lolmaus:
СС @kytwb.

kytwb:
@lolmaus Sorry for the inconvenience you experienced with this upgrade 🙏 I originally proceeded for https://github.com/getferdi/ferdi/issues/300 as the context menu was not implemented in the previous Telegram client. Linking to https://github.com/evgeny-nadymov/telegram-react/issues/116 to reference one of the issue you experienced with this new client. I will take a look at reverting and offering this client as a new recipe as per @vantezzen suggestion.

lolmaus:
Thx!

I didn't see lack of right click menu as an issue because all the features that Telegram-React offers were available in original Telegram-Web via left click. 🤷‍♂️

kytwb:
I see; I didn't know, I don't use Telegram that much, so thank you for the clarification 🙏 I will proceed as discussed then.

PR #132, opened 2020-04-21
Switch back to original Telegram, add Telegram React

kytwb:
For #128 and https://github.com/getferdi/ferdi/issues/623.


kytwb:
Screen Shot 2020-04-21 at 10 18 02 PM

Issue #57, opened 2020-03-04
Update GitHub to count amount of notifications


kytwb:
**Is your feature request related to a problem? Please describe.**
Right now the GitHub recipe doesn't leverage notifications.

**Describe the solution you'd like**
GitHub shipped a nice notifications center (in beta) at https://github.com/notifications/beta.

<img width="838" alt="Screen Shot 2020-03-04 at 1 10 27 AM" src="https://user-images.githubusercontent.com/412895/75831862-f5f21180-5db4-11ea-974c-1b0b77a67de8.png">

@vantezzen do you think it would make sense to update the recipe to handle notifications? I think it could be useful, and we could start [turning off GitHub email notifications](https://twitter.com/arcanis/status/1234916836577288192) completely, eventually 😁 
WDYT?

vantezzen:
Sounds like a good idea to add notification support

FeikoJoosten:
There you go. #129

kytwb:
@FeikoJoosten Thank you so much! 💖 Keeping this issue open so we remember to implement proper notifications count later on! 🙏

kytwb:
Pending global availability of notifications dashboard that includes notifications count (as per screenshot in original message).

kytwb:
Just doubled-check and both versions have the unread notifications count:

<img width="634" alt="Screen Shot 2020-04-21 at 10 52 13 PM" src="https://user-images.githubusercontent.com/412895/79922083-ccec1580-8422-11ea-91ea-f7641a2e8db1.png">
<img width="287" alt="Screen Shot 2020-04-21 at 10 52 23 PM" src="https://user-images.githubusercontent.com/412895/79922084-ceb5d900-8422-11ea-9a27-746db672fa93.png">

kytwb:
In both case it seems the `.count` selector matches the right notification count.

kytwb:
That would involve changing the service URL to https://github.com/notifications though. What do you think?

FeikoJoosten:
Just wondering, what does the service URL do exactly. Is that the url that'll be opened by Ferdi? Or is it the url that's used for checking notifications?

kytwb:
@FeikoJoosten It's the URL that is loaded when Ferdi loads the service. To check the notifications, we don't use an URL, we continuously check the current page of the service for notifications indicators in the DOM of the page (direct count in notifications center or indirect blue dot in the header in the case of GitHub).

FeikoJoosten:
[...]

PR #133, opened 2020-04-22
Use GitHub notifications center and direct notifications

kytwb:
For https://github.com/getferdi/recipes/issues/57.

The service now starts on a login screen instead of GitHub landing:

Screen Shot 2020-04-22 at 1 18 20 AM

Shows direct notifications count on both old/new GitHub notifications center:

Screen Shot 2020-04-22 at 1 05 04 AM

Screen Shot 2020-04-22 at 1 12 40 AM

If user navigates away from notifications center, we display an indirect notification based on GitHub header notifications indicator:

Screen Shot 2020-04-22 at 1 12 06 AM

Also fixes the GitHub recipe as I let something slip through during the review (see https://github.com/getferdi/recipes/pull/129/files#r412590349).


uncompressed/github/webview.js
```diff
@@ -2,9 +2,19 @@
...
   const getMessages = function getMessages() {
-    Franz.setBadge(document.querySelector('[class*="mail-status unread"]') !== null ? 1 : 0);
+    const directCountElement = document.querySelector('.filter-list .count');
+    const indirectCountElement = document.querySelector('[class*="mail-status unread"]')
+    let directCount, indirectCount
+    if (directCountElement) {
+      directCount = parseInt(directCountElement.innerHTML, 10);
+    } else {
+      if (indirectCountElement) {
+        indirectCount = 1;
```
FeikoJoosten, line 12: You can always add the indirect count. The direct count is taken with a priority. But for now it doesn't matter that much. I don't think at the moment there are any UI reasons to always include both.

kytwb: @FeikoJoosten Noted!  
I didn't remind exactly how setBadge worked so, in doubt, I played it safe 😄

FeikoJoosten: Hehe can understand. Better safe than sorry yes! It is something to keep in mind though. If there's ever support to view both direct and indirect notifications at the same time.

PR #136, opened 2020-04-24
Update Integration documentation

mahadevans87:
- Update the possibility of overriding user-agent in ferdi recipes so that user's know this options is possible.


vantezzen:
Thank you for this addition. I'll extend this section a little bit more after merging

PR #140, opened 2020-04-26
WhatsApp - Chrome error every time the service is loaded

mahadevans87:
This issue addresses the issue with whatsapp not loading with the default electron user-agent during Ferdi start or service load. For more info refer https://github.com/getferdi/ferdi/issues/603

**Description**

replace overrideUserAgent with modifyRequestHeaders Hook and pass the user-agent header inside this function. This is because overrideUserAgent is called after the service is loaded and hence whatsapp loads with the default user agent.

Note: Passing the user-agent header in modifyRequestHeaders() ensures that all the request headers are sent before the page is loaded. We might need to use this hook in other services as well.

**Screenshots**

ferdi_603_fix


mahadevans87:
@vantezzen @kytwb - Can you please review this?

kytwb:
@sampathBlam I agree! 😄

PR #142, opened 2020-04-27
Pasting text from clipboard into hangouts prepends it

mahadevans87:
Pasting text from clipboard into hangouts prepends it to the already existing message
  - replace hardcoded user-agent with the user-agent used in other recipes
  - replace overrideUserAgent() to modifyRequestHeaders() hook.
ferdi_643_fix

Fixes https://github.com/getferdi/ferdi/issues/643
Co-authored-by: Sampath Kumar Krishnan <sampathblam@users.noreply.github.com>


mahadevans87:
@kytwb can you please review this?

Issue #141, opened 2020-04-26
Jitsi app doesn't work on Ferdi 5.5.0-gm.2


tristanplouz:
**Describe the bug**
The service doesn't work anymore with Ferdi 5.5.0-gm.2, the screen of the service still gray.

**To Reproduce**
Just try to start the service

**Expected behavior**
The website view is expected

**Screenshots**

**Desktop (please complete the following information):**
 - OS:Linux
 - Version 5.5.0-gm.2

**Additional context**

cybermoloch:
#146 was incomplete; #167 fixes this with the proposed changes as per @sampathBlam.

kytwb:
The fix has been merged and will ship in the next release of Ferdi.

PR #115, opened 2020-04-19
Fix user agents override methods

kytwb:
For https://github.com/getferdi/recipes/issues/114, https://github.com/getferdi/ferdi/issues/536, https://github.com/getferdi/ferdi/issues/603, https://github.com/getferdi/ferdi/issues/598, https://github.com/getferdi/ferdi/issues/601, and potentially https://github.com/getferdi/ferdi/issues/475.

PR #146, opened 2020-04-28
Correction on the userAgent definition to work with the Jitsi website

tristanplouz:
Correction for issue #141


uncompressed/jitsi/index.js
```diff
@@ -1,2 +1,6 @@
 "use strict";
-module.exports = Ferdi => Ferdi;
\ No newline at end of file
+module.exports = Franz => class Jitsi extends Franz {
+  overrideUserAgent() {
+    return window.navigator.userAgent.replace("Electron ","Electron/");
```
sampathBlam, line 4: Can you check if the following user agent works? This is what has been followed in most of the other recipes and it would be great if it works so that we can maintain uniformity among recipes. 
(Reference: #115)

```suggestion
    return window.navigator.userAgent.replace(/(Ferdi|Electron)\/\S+ \([^)]+\)/g, '').trim();
```

tristanplouz: Hello,
It works but the detected browser is Chrome and not Electron, I don't know if it will create some issues after.

sampathBlam: @tristanplouz: That shouldn't be a problem. Whatsapp uses the same code and the user-agent that is being sent from my test machine is 
```
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.163 Safari/537.36 Ferdi/5.5.0 (Electron 8.2.3)
```
Just trying to maintain uniformity here.

PR #98, opened 2020-04-13
Updated nextcloud talk icons for 1.0.1

JakeSteam:
[Issue raised](https://github.com/getferdi/recipes/issues/96) about incorrectly using Messenger icons. 
Updated with icons from [Nextcloud trademarks page](https://nextcloud.com/trademarks/) (download required).

image

PR #150, opened 2020-05-01
Adds Guilded.gg recipe

JakeSteam:
Adds support for [Guilded.gg](https://www.guilded.gg/). Currently only supports the current server.

Uses assets from official [press kit](https://www.guilded.gg/brand).


JakeSteam:
Looks like the same Windows issue as #98 :(

kytwb:
@JakeSteam Thank you for your contribution and sorry for the inconvenience with the CI 🙏

PR #153, opened 2020-05-06
Add Clubhouse

dnlnrs:
## Adding new recipe
Service: Clubhouse

URL: https://clubhouse.io

Service ID: teamId

### Terminal output
```
[Info] Creating temporary directory
[Info] Downloading https://github.com/dnlnrs/ferdi-recipe-clubhouse
[Info] Decompressing repository
[Info] Moving 'recipe_src' to 'recipe_src_tmp'
[Info] Packaging your recipe
[Info] Packaging clubhouse...
[Info] Copying to uncompressed recipes
[Info] Adding to all.json
[Info] No recipe with ID found - creating new.
✅ Successfully packaged and added new recipe "clubhouse"
[Info] Deleting temporarydownloaded repository
[Info] Moving back 'recipe_src_tmp' to 'recipe_src'
✅ Successfully packaged the recipe from your GitHub repository
```

### Additional information

- [x] I am the original creator of this package
- [X] My recipe has been tested to work inside Ferdi


uncompressed/clubhouse/webview.js
```diff
@@ -0,0 +1,11 @@
+module.exports = (Ferdi) => {
+  function getMessages() {
+    let unread = 0;
+    const hasNotifications = document.querySelector('#notifications-link .badge').classList.contains('visible');
+    if (hasNotifications) {
+      Ferdi.setBadge(1);
```
mahadevans87, line 6: Thanks for the contrib. Any particular reason why the notification count is  set to a constant value (1) ?

dnlnrs: Because unfortunately there's no a number but only a red badge. If you prefer I can completely remove the badge check.

image

mahadevans87: Thanks for the clarification. Can we classify this as an indirect notification then? There is an API that you can use to set this as an indirect notification which would not show any number
`    Franz.setBadge(directCount, indirectCount);
`

dnlnrs: Sure! 👍

PR #154, opened 2020-05-06
Adds Clubhouse recipe

janhohner:
Adds a recipe for https://clubhouse.io


janhohner:
Just saw #153

dnlnrs:
We had the same idea the same day 😄

PR #156, opened 2020-05-06
Updated Outlook/OWA logos

cybermoloch:
Updated logos for Outlook/OWA.


cybermoloch:
Can't repackage office365-owa. It doesn't like the numbers in the id:
`"Your package.json defines an invalid recipe ID. Please make sure the 'id' field only contains lowercase letters (a-z) and hyphens (-)"`

cybermoloch:
Made PR 158 to fix the regex for numbers in id check. PR 157 includes both this and the previous PR for updating the logos.

As a note, it would be kind of nice to allow editing of the uncompressed files and have GitHub Actions do the packaging so it would be easier for quick edits.

kytwb:

I fixed the conflict via https://github.com/getferdi/recipes/commit/02e25c48693c2d2a72d54bdf4204e733949393c1 but wasn't able to give credit/mark the commit as co-authored, apologies for that.

I will be closing this pull request and overlook the archiving issue as there is an on-going refactor via https://github.com/getferdi/recipes/pull/292.

PR #158, opened 2020-05-08
Fix for packaging not allowing numbers in recipe ids

cybermoloch:
Fix for recipe id validity check to allow numbers as at least one existing recipe had numbers in the id but could not update because of this. The regex also allowed periods and underscores so updated the error messages as appropriate.

This is under the assumption that numbers, periods and underscores SHOULD be valid ids.

### DIFF
--- a/scripts/api/package.js
+++ b/scripts/api/package.js
@@ -93,9 +93,9 @@ For more information about the package.json file visit https://github.com/getfer
   }
   let configErrors = [];
   if (!config.id) {
-    configErrors.push("Your package.json contains no 'id' field. This field should contain a unique ID made of lowercase letters (a-z), numbers (0-9), hyphens (-), periods (.), and underscores (_));
+    configErrors.push("Your package.json contains no 'id' field. This field should contain a unique ID made of lowercase letters (a-z), numbers (0-9), hyphens (-), periods (.), and underscores (_)");
   } else if (!/^[a-z0-9._\-]+$/.test(config.id)) {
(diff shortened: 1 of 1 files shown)


cybermoloch:
Sorry, not good at keeping PRs clean.

Question about SVG: The size shouldn't matter as long as they are square but I "resized" these anyway because somewhere it said they should be 1024x1024. Should I resubmit the original SVGs before I "resized" them. The SVG size check was also removed in a previous commit -- should it be reinstated to at least check the width and height dimensions are the same to ensure a proper square?

sampathBlam:
@cybermoloch : Can we have only the changes with respect to the packaging flow in this PR? I understand that the recipe ID limitation is preventing you from updating some recipes. But moving the recipe updates to a separate PR is better. And this PR could be referred in future only for packaging changes. 

I believe there should have been some reasoning behind restricting recipe IDs to contain only lowercase letters and hyphens. Or it could have been followed like how it is in Franz. Probably @kytwb or @vantezzen can provide more info regarding this.

As far as SVGs are concerned, if it is not gonna make it tougher for contributors to update recipes (because we don't get square images for logos that easily), it is good to have the check to identify if the SVG is a square image.

cybermoloch:
Fixed the PR and made another for the SVG check in case it is warranted.

sampathBlam:
@vantezzen : If there is no particular reason to disallow numbers in recipe id, can we proceed merging this PR ?

PR #161, opened 2020-05-09
Add check to ensure SVG icon is a square

cybermoloch:
Assuming that SVG icons should be a square so they don't get deformed if they are resized without maintaining aspect ratio.

SVGs that I checked (half dozen or so) were already squares but did not check the Ferdi code that displays the icons to ensure that it will not deform them if they are not squared.


cybermoloch:
Ah, okay. I only did a test with one or the other. (I did use the previous removal as reference but thought they could be separate.) I will do a fix.

cybermoloch:
Thinking about this further, should all errors be checked and output together in that case? All checks exit with the one error it found (checking for valid id, png/svg, etc).

cybermoloch:
Used the idea from the config JSON check and changed all errors to build an array, outputs the full array with all error messages or continues if the array is empty.

I would like to clean this up more and also make the config one use the same array per check instead of its own if this is an acceptable method for handling the packaging errors.

cybermoloch:
Fixed up the formatting and error messages. Also folded the JSON validation into the same array as the other error messages.

I think this satisfies your request and then makes it work the same way for all current validation as well. (They just exited before except for the icons which were together.) Also, the errors output can be quite lengthy and hard to read as they didn't have good separation (since many were multiple lines or long enough to wrap), so I added some better formatting to help separate the errors.

kytwb:

I am linking https://github.com/getferdi/recipes/pull/292 for reference as this may conflict with the on-going refactor.

PR #162, opened 2020-05-09
Add Gotomeeting recipe

tristanplouz:
Add a new service: Gotomeeting


uncompressed/gotomeeting/webview.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
+
+const path = require('path');
+
+module.exports = Ferdi => {
+  Franz.injectCSS(path.join(__dirname, 'service.css'));
```
vantezzen, line 6: Looks like you have a bug in your code here: I get a "Franz is not defined" error, this should probably be "Ferdi".
Please keep in mind that you'll have to **reversion and repackage** your recipe after changing this.

kytwb:
```suggestion
  Ferdi.injectCSS(path.join(__dirname, 'service.css'));
```

kytwb:
@tristanplouz thank you for your contribution!

Issue #149, opened 2020-05-01
Further enhancement for "Jitsi Meet" on your application !


Nokia808:
Hi dear developers.

I discover - while trying "jitsi-meet-electron" package (AppImage) a feature I'm sure that you did not add it when you add Jitsi Meet to your application as I requested from you .....

The issue is that: Jitsi is MULTI-INSTANCES ! This mean that every one can build heir/his own Jitsi instance (server) & not obligated to use https://meet.jit.si which is official instance of the developer teams of Jitsi .... For example:

- we have www.infomaniak.com added it's own Jitsi Meet (which working on their own server not related to Jitsi developers team. See the following links on their web site:
https://www.infomaniak.com/en/meet
https://meet.infomaniak.com/?lang=en
The infomaniak own Jitsi Meet may has more capacity than server of https://meet.jit.si because Infomaniak is a big company .....

- other example is disroot.org own Jitsi - see:
https://disroot.org/
https://calls.disroot.org/

I think when you implement Jitsi Meet you considered https://meet.jit.si only, isn't it ?

You should add a feature in setting of Jitsi Meet on Ferdi allow user to change https://meet.jit.si to what she/he wish like for example Infomaniak.com own Jitsi Meet .....

Here is a screenshot from setting of jitsi-meet-electron that I already downloaded it & started to use it just today:
![jitsi-meet-setting](https://user-images.githubusercontent.com/25375330/80833509-32d45c00-8bde-11ea-9e42-75740c24115d.png)

Look to screenshot carefully & you will see a field [Srver URL] contain "https://meet.jit.si" , which is EDITABLE field .....

You can download "jitsi-meet-electron" package from:
https://github.com/jitsi/jitsi-meet-electron/releases

Thank you.

cybermoloch:
PR #167 Fixes this

Nokia808:
@cybermoloch 
Really amazing work ! Thank you too much !

kytwb:
This will be available in the next release of Ferdi.

PR #167, opened 2020-05-10
Update Jitsi Meet to allow self-hosted versions and UserAgent fix

cybermoloch:
Updated Jitsi recipe to have hasCustomUrl: true. Left the default server in the serivceURL for a future version that may fill in that as a default server but would allow editing if 'hasCustomUrl' was true.


cybermoloch:
Some notes here:

This should actually be Jitsi Meet as they have other "Jitsi" projects (Jitsi Desktop in the past so that comes to mind first when I hear Jitsi).

Leaving the serviceURL in there even with hasCustomUrl: true is necessary in this case of a switch over because otherwise Ferdi will just show "Loading Jitsi" forever with it switching. This is behavior should probably be documented in the recipe change/update docs if that is intended. (However, it does not show up in the configuration tab and will still insist it cannot be empty.)

Nokia808:
Is there a hope to fix this & merge it ?

all.json
```diff
@@ -444,7 +444,7 @@
     "featured": false,
     "id": "jitsi",
     "name": "Jitsi",
```
kytwb, line 446:
```suggestion
    "name": "Jitsi Meet",
```

uncompressed/jitsi/package.json
```diff
@@ -1,13 +1,15 @@
 {
   "id": "jitsi",
   "name": "Jitsi",
```
kytwb, line 3:
```suggestion
  "name": "Jitsi Meet",
```

uncompressed/jitsi/package.json
```diff
@@ -1,13 +1,15 @@
...
-  "version": "1.0.2",
+  "version": "1.0.3",
   "description": "Jitsi",
```
kytwb, line 5:
```suggestion
  "description": "Jitsi Meet",
```

kytwb:
I also updated the name of the recipe to "Jitsi Meet" as per your suggestion.

I merged despite the failing build as part of the refactor taking place in #292.

kytwb:
Fixes https://github.com/getferdi/recipes/issues/149.

PR #168, opened 2020-05-11
Franz -> Ferdi

obvionaoe:
Switching from Franz to Ferdi in the recipe README's

Just did one, but I can do the rest if you approve this change!

### DIFF
--- a/uncompressed/NewsBlur/README.md
+++ b/uncompressed/NewsBlur/README.md
@@ -1,5 +1,5 @@
-# NewsBlur for Franz
-This is the Franz recipe for NewsBlur
+# NewsBlur for Ferdi
+This is the Ferdi recipe for NewsBlur
 
-### How to create your own Franz recipes:
-* [Read the documentation](https://github.com/meetfranz/plugins)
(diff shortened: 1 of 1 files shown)


vantezzen:
First off, thank you for the initiative.
I don't think this is necessary and could actually hinder us in keeping the recipes up-to-date:
After you change a file in the recipes you'll have to repackage and reversion it (see our docs at <https://github.com/getferdi/recipes/blob/master/docs/updating.md>). If we don't do that, the recipe repository won't be valid: you can already see the check is already failing on this PR as you changed a file without repackaging and reversioning it.
After reversioning, we'll not have the same version numbers that the original recipe creators use in their repositories.

The goal for the future would be to automatically keep recipes up-to-date with their original repositories - if we reversion all recipes, this check won't work reliably. Additionally, if a recipe updates, it would also rewrite all texts with "Franz" again.

We could however keep your change in mind and automatically replace "Franz" with "Ferdi" in that updater script before we package the recipe - this way no reversioning is needed as we haven't packaged the new version yet.

PR #169, opened 2020-05-13
Enterprise OWA-Custom URL

tfehren:
Added a recipe for Enterprise installations of Outlook Web App.

- The URL can now be specified.

Could also update or change the other OWA recipe to cater to all use cases.


vantezzen:
Thank you for your contribution! I think it's a good idea to have this in a separate recipe as we currently cannot define a default value for the server.

Please keep in mind for future PRs that recipes need to be packaged - otherwise Ferdi won't detect them (see https://github.com/getferdi/recipes/blob/master/docs/integration.md#publishing). If you are unfamiliar or uncomfortable with packaging recipes we can also package them for you, but it would save us work if you already package it on your side.

vraravam:
Can I request you to please delete the original PR branch since its been merged? Otherwise, it causes some "noise" in the forked repos as well. Thanks in advance.

PR #189, opened 2020-05-30
Big Blue Button - Custom URL recipe

tfehren:
New Recipe for BigBlueButton: https://bigbluebutton.org/


vantezzen:
Thank you for your contribution! Why did you name your recipe "BigBlueButton Enterprise" instead of just "BigBlueButton"?

tfehren:
I am open to changing the name:
As I'm using it with our Company BigBlueButton Server. Not sure if there is a public one... As you have to specify your own server it is named it Enterprise as it does not work without specifying a server.

vantezzen:
BigBlueButton is completely self-hosted (apart from the Demo server but that doesn't really count) that's why I suggest changing it to just "BigBlueButton" as the Enterprise might confuse people to if the recipe is only for specific servers.

tfehren:
@vantezzen done

vantezzen:
Great, thank you for your contribution!

PR #194, opened 2020-06-01
Fix popup blocker error message while opening links in Gmail

mahadevans87:
Fix popup blocker error message while opening links in Gmail
 - Gmail does not like custom user agents (in fact, any of the google apps)
 - replace user-agent in gmail by removing any Ferdi|Electron strings and trim() the end result

Screenshots
gmail_grr


all.json
```diff
@@ -254,10 +254,10 @@
   },
   {
     "author": "Stefan Malzner <stefan@adlk.io>",
-    "featured": true,
+    "featured": false,
```
mahadevans87, line 257: Not sure why this featured flag is inverted! @sampathBlam @vantezzen any ideas why?

sampathBlam: Apparently the [package script](https://github.com/getferdi/recipes/blob/master/scripts/api/package.js) is setting the value of `featured` to false always. Unlike other parameters like author, version etc, it is not copied from the package.json. I do not see any documentation regarding this flag in the package.json.
Digging further, I see that the value was changed manually in this PR (#151) last time. 
@vantezzen : What is the significance of this featured flag ?

vantezzen: Yes, the script currently doesn't pay attention to the original value of `featured`.

The flag is used to show these recipes on the "Most used services" page when adding new services.

mahadevans87: So Do we leave it like this? Or shall change the featured flag back to true?

vantezzen: The featured flag should be set to true - manually for now and I'll look into implementing that into the packaging script in the future.

mahadevans87:
@vantezzen made the changes. Requesting you to merge if everything looks good.

vantezzen:
Great work! I just repackaged and merged.

mahadevans87:
@vantezzen It seems this fix never went into the updated recipes. Can you confirm? I just downloaded a brand new version and it still had the older piece of code. Can you please check and we can close - https://github.com/getferdi/ferdi/issues/321 ?

jottr:
@vantezzen please respond to @mahadevans87 question. It's been a month now.
I'd like to see this bug fixed.

jottr:
How is this merged in the first place? The integration test failed.

vantezzen:
> It seems this fix never went into the updated recipes

As far as I can tell the changes are merged into the codebase. This PR only seemed to have changed the way the header gets generated and I can see that change in https://github.com/getferdi/recipes/blob/master/recipes/gmail/index.js.

> Why has this been merged in the first place? The integration test failed

This was because of an incorrect package. As I wrote above, I repackaged and the tests never re-ran after that. As you can see at https://github.com/getferdi/recipes/commit/3528aeee325312ec1668cb614ec3e4e1424b7f99, the tests didn't fail after merging.

jottr:
Thank you @vantezzen. 
You can manually trigger a build in case it doesn't trigger for some reason or other.
Did you confirm that the patch actually works before merging?

vantezzen:
> You can manually trigger a build in case it doesn't trigger for some reason or other

It did trigger but we didn't needed it because I tested it locally before pushing and the CI build doesn't do anything else.

> Did you confirm that the patch actually works before merging?

Yes it works

chauhanankit83:
I just found out the below code to fix the popup blocker issue.

var os = require('os')

module.exports = Franz =>
  class Gmail extends Franz {
    overrideUserAgent() {
      if (os.platform() == 'linux')
        return "Mozilla/5.0 (X11; Linux x86_64; rv:72.0) Gecko/20100101 Firefox/72.0"
      else
        return "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.89 Safari/537.36"
        // original
        // return "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:72.0) Gecko/20100101 Firefox/72.0";
    }
  };

onechaiplease:
I just found out the below code to fix the popup blocker issue.

var os = require('os')

module.exports = Franz =>
class Gmail extends Franz {
overrideUserAgent() {
if (os.platform() == 'linux')
return "Mozilla/5.0 (X11; Linux x86_64; rv:72.0) Gecko/20100101 Firefox/72.0"
else
return "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/84.0.4147.89 Safari/537.36"
// original
// return "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:72.0) Gecko/20100101 Firefox/72.0";
}
};

PR #209, opened 2020-06-17
Fix WhatsApp fullscreen

breuerfelix:
WhatsApp is not in fullscreen mode.

At least for me, the container does not have a css class app. It is called 'two'.

### DIFF
--- a/uncompressed/whatsapp/service.css
+++ b/uncompressed/whatsapp/service.css
@@ -1,4 +1,4 @@
-.app-wrapper-web .app {
+.app-wrapper-web .two {
     width: 100% !important;
     height: 100% !important;
     top: 0 !important;


breuerfelix:
i dont really understand what i need to change in order for the test to pass.

do i need to upload a new archive? if yes, what is the point of having the code packaged and unpackaged in the repository ? :D

breuerfelix:
@vantezzen okay thanks for the explanation.

before doing the work, i havent found any issues regarding whatsapp so i ask myself if i am just the first person encountering the issue or is it just me having problems (i will change my user.css then)

is your whatsapp also not fullscreen ?

vantezzen:
I haven't had any issues with WhatsApp lately. Maybe its best if you use a css selector like `.app-wrapper-web .app, .app-wrapper-web .two` instead, so your issue gets fixed but that it doesn't create any new issues for other users.

breuerfelix:
i just figured out when this happens.

if you click on a user profile to see the the image then the windows will pop out of fullscreen.  
if you close the user profile again, it will be fullscreen again :)

i will hopefully submit a proper PR during this week

PR #216, opened 2020-06-23
Fix whatsapp fullscreen for different screen sizes

breuerfelix:
title says it all :)

finally with a correct PR

depending on the screen width, the css class is changing. and also depending on if you open the profile image from a contact

Signed-off-by: Felix Breuer <f.breuer94@gmail.com>


vantezzen:

PR #229, opened 2020-07-07
Fixing unread-counter for office365

CrEaK:
Now with a correct PR :-)
The unread counter was not working for me. I fixed it. Seams the html-structure has changed.


vantezzen:
Thank you for this contribution, too 😄! Again, your changes will be shipped in the next version of Ferdi.

PR #230, opened 2020-07-07
Mattermost: Fix badge for unread channels when in single team

CrEaK:
The unread notification badge was not shown in mattermost, when you only have unread channels. This seams to happen, when you are in a single team and there is no team-switcher visible.

With my simple change there is now a unread notification badge :-)


vantezzen:
Thank you for your contribution! Your changes will be shipped in the next version of Ferdi.

PR #235, opened 2020-07-15
Update Riot.im to Element.io

omove:
Riot.im has changed their name to Element.


vantezzen:
Thank you for your contribution! Your changes will be shipped with the next version of Ferdi.

PR #236, opened 2020-07-17
Add support for empty unread badges in WhatsApp

ruippeixotog:
Empty unread badges show up when people manually "Mark as Unread" chat threads. Currently, this stops notifications from appearing as counts become NaN. With this PR they will stop being counted (since we don't have message counts in those cases and it's not really a notification I think this is probably the best behavior, but I'm ok with defaulting to 1).


vantezzen:
Thank you for your contribution.
Please follow our guide at <https://github.com/getferdi/recipes/blob/master/docs/updating.md> in order to update. This repository contains the unpackages files only for convenience purposes , but they are not actually used by Ferdi, so you'll have to repackage the recipe.

ruippeixotog:
@vantezzen sorry about that, I should have read he instructions first 🙁 

It should be good to merge now.

vantezzen:
No problem, thank you for packaging it and thank you again for your contribution! Your changes will be released with the next release of Ferdi.

PR #244, opened 2020-07-26
Recipe for Carnet (hosted on a Nextcloud server)

eandersons:


eandersons:
New pull request: #263 

It seems that the Travis CI check does not pass if the recipe was packaged on Windows.

PR #245, opened 2020-07-26
Recipe for Nextcloud Cospend

eandersons:
Nextcloud Cospend is a group/shared budget manager.


eandersons:
New pull request: #262 

It seems, that the Travis CI check does not pass if the recipe was packaged on Windows.

Issue #233, opened 2020-07-10
[Recipe Request] Feedly


gustavohellwig:
Name of the service: Feedly
URL of the service: https://feedly.com
GitHub URL to a recipe: https://github.com/feedly/

kytwb:
This will be available in the next release of Ferdi.

PR #254, opened 2020-08-07
Add Feedly recipe

kastnerp:


uncompressed/feedly/webview.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const path = require('path');
+
+module.exports = Franz => {
+  let modal;
+  let updates = 0;
+
+  const createModal = () => {
```
kytwb, line 9: I don't understand why a modal is being created.

kytwb (changes requested):
I see there is a Franz recipe available at https://github.com/alphamyd/meetfranz-recipe-feedly, have you tried it?

uncompressed/feedly/webview.js
```diff
@@ -0,0 +1,41 @@
...
+
+  const hideModal = () => {
+    modal.querySelector('p').innerHTML = '';
+    updates--;
+    modal.classList.remove('open');
+  };
+
+  window.alert = showModal;
+
+  const getMessages = () => Franz.setBadge(updates);
```
kytwb, line 32: I don't use Feedly so I am not sure if it is really needed to override `window.alert` with this modal logic. Isn't it possible to get the notifications count thanks to a call to `document.querySelector` like in https://github.com/alphamyd/meetfranz-recipe-feedly/blob/6b5d39962d516dd4b6cfdd74526d310f6cf321a5/webview.js#L4?

kastnerp:
You are right, I had not tested the other one.

uncompressed/feedly/feedly.css
```diff
@@ -0,0 +1,41 @@
...
+    font-size: 20px;
+    font-weight: bold;
+}
+
+#franz-modal .close:hover,
+#franz-modal .close:focus {
+    color: black;
+    text-decoration: none;
+    cursor: pointer;
+}
```
kytwb, line 41:
```suggestion
```

kytwb:
@kastnerp thank you for your contribution!

I merged despite the failing build as part of the refactor taking place in #292.

kytwb:
Fixes https://github.com/getferdi/recipes/issues/233.

PR #261, opened 2020-08-18
Recipe for Nextcloud Tasks

eandersons:


vantezzen:
Thank you for your contribution! Your recipe will be shipped with the next version of Ferdi.

PR #262, opened 2020-08-18
Recipe for Nextcloud Cospend

eandersons:
Nextcloud Cospend is a group/shared budget manager.


vantezzen:
Thank you for your contribution! Your recipe will be shipped with the next version of Ferdi.

PR #264, opened 2020-08-18
Add Nextcloud Calendar recipe

eandersons:
The Calendar app is a user interface for Nextcloud's CalDAV server.


kytwb:
@eandersons thank you for your contribution!

armaccloud:
Hi @kytwb @eandersons 
I am unsure where to report this, but in my set-up the calendar isn't loading the middle (calendar) part.
Do you have any idea how I can troubleshoot this?

image

eandersons:
@armaccloud, you could try to open Developer Tools (View > Toggle Service Developer Tools) and look for seemingly important resources that are not loaded in the Network tab or inspect errors in the Console tab (there are some errors even for a working Nextcloud Calendar service).

Another option would be to add the Nextcloud service to Ferdi and open calendar to see if it is working there. If it is not loading in the Nextcloud service either, I would guess that some requests are blocked by the server. Then you should look at web server (nginx, Apache etc.) and possibly Nextcloud logs.

armaccloud:
@eandersons 

Thank you for these troubleshooting steps!
I added Nextcloud and verified that the calendar is working there correctly.

When showing the console logs for the Calendar Recipe, it shows me the following.

image
image


Maybe "refuse to execute inline script" and "pauserendering" have something to do with it?

eandersons:
@armaccloud, seems that the culprit could be some or all of the following errors:

- calendar: First day parameter contains non-numerical components
- Error: viewType "apps" is not available
- TypeError: Cannot read properties of undefined
- Failed to load resource: the server responded with a status of 404 (Not Found)

Unfortunately I am not sure how to troubleshoot those. You could open the Network tab and then reload the Nextcloud Calendar service to see if there are resources that failed to load.

Just in case, URL in the service settings field "Custom server" is something like this: <https://nextcloud.example.org>?

Did you try to remove and re-add Nextcloud Calendar service?

armaccloud:
@eandersons 
Thank you for looking into this with me.
I found the issue; in the custom server, I had to put index.php behind the URL, so

https://{nextcloud}/apps/calendar/
actually has to be 
https://{nextcloud}/apps/calendar/index.php

Now everything works :)

PR #268, opened 2020-08-22
Add Campuswire recipe

Ninja73737:
I based these modifications on the android messages and messenger recipes. The webview.js is the only thing I'm not completely sure is correct cause I don't completely understand it, but it seems to be working for me.


kytwb:
@Ninja73737 thank you for your contribution!

Ninja73737:
Ok, thanks @kytwb ! I also created a Tor Check recipe with https://check.torproject.org so that people can test their tor proxies out to confirm they are properly set up, should I wait until after the refactor and add the recipe in then, or should I create a PR now?

kytwb:
@Ninja73737 I don't think it makes sense to have a recipe just to have access to this check. If it does for you, a Custom Website service is more appropriated. 

Nonetheless, I think this could implemented as an enhancement on the existing proxy settings. You are welcome to create a new issue to propose an enhancement on the main repository with some more details on the goal(s) we want to reach (do we, for example, want to be able to restrict usage of services when, and only when, connected to the Tor network?).

Ninja73737:
@kytwb Oh sorry, that's my bad, I totally didn't realize that the custom website was a thing...
I think the current behaviour for proxies is fine, at least for the way I have mine set up. I'm running polipo as an http proxy, then running polipo through tor, so when the proxy isn't connected it just gives a proxy connection error and doesn't load the service,  I think it should behave this way for most users, so it may not be necessary to add that.

PR #270, opened 2020-08-25
Update Nextcloud Talk to remove/disable unnecessary links

eandersons:
`service.css` was updated to remove or disable unnecessary links from the top
menu to remove the possibility to navigate away from the app:

- disable Nextcloud logo, Talk and settings link;
- remove "More" and contacts link.

`package.json` was updated to:

- increase version number to 1.0.2;
- add `serviceURL` to `config` to successfully package this recipe.


kytwb:
@eandersons thank you for your contribution!

I don't use Nextcloud Talk personally and wondered if those changes were necessary as they seem quite opiniated. Some users browse within Ferdi as far as I know.

I merged despite the failing build as part of the refactor taking place in #292 and trust your judgement as an user of this recipe.

eandersons:
@kytwb, well, it depends on the point of view. As I see it, Ferdi is a wrapper for individual services and therefore if there is a service named Nextcloud Talk then an user shoud not navigate away from it as it may be a little confusing - the service is named Nextcloud Talk, but it contains, lets say, Nextcloud Files. If there would be a service named Nextcloud, then yes - nothing from navigation elements should be hidden or disabled and the user should be able to freely navigate within Nextcloud.

kytwb:
Thank you for sharing your thought process, it makes sense to me as well!

PR #271, opened 2020-08-26
Fix wrong notifications count for Slack

FeikoJoosten:
This PR fixes an issue where the `All Dms` was counted as a valid notification causing an additional increase in the unread messages count.


kytwb:
@FeikoJoosten thank you for your contribution!

PR #292, opened 2020-09-19
Unpack recipes and update recipes icons

vantezzen:
Adding a recipe to Ferdi is hard - way harder than it needs to be! You can't just update some text files inside this repository and be done - instead you'll have to worry about "packaging" the recipe. Even after that, many developers on Windows seem to have problems with our packaging system in that packaged recipes don't correctly work.

This PR will move the whole package chain to our Ferdi build. With this, developers won't have to package their recipe themselves, instead we package all recipes at once during Ferdi's normal build process.

All recipes are now stored inside the `recipes/` folders and are build into the `archives/` folder as before. The script will also dynamically create the central `all.json` file and populate it with the latest information.

The necessary changes to the main Ferdi project will be made with https://github.com/getferdi/ferdi/pull/966


kytwb:
@vantezzen it looks good to me!

We would need to merge this into master for the build on getferdi/ferdi#966 to turn green.

How do you suggest we proceed to deal with the open pull requests? I propose reviewing them and merging them, then syncing your pull request with master and re-applying your edits to the new/updated recipes.

What do you think?

vantezzen:
Yes, that sounds good - most of the PRs aren't merged yet because they needed repackaging but that can be ignored now

kytwb:
Current status:

```
kytwb@kytwbs-mbp recipes % npm run package

> ferdi-recipes@1.0.0 package /Users/kytwb/Projects/Ferdi/ferdi/recipes
> node scripts/package.js

⚠️ Couldn't package "sendtask": Recipe doesn't contain an icon PNG
⚠️ Couldn't package "yammer": Recipe doesn't contain an icon PNG
✅ Successfully packaged and added 123 recipes (2 unsuccessful recipes)
```

I think the command should fail (exit code != 0) if one or more recipes couldn't be packaged, what do you think?

Side notes, I couldn't properly merge back the icon size check introduced by #161, the change got lost in the process. I also think we can remove the `meetfranz-` prefix from the LastPass recipe identifier to harmonize.

kytwb:
Another note on this; the archives were useful for recipes "micro-updates". As far as I remember, Franz was able to update a recipe remotely (showing a "your service has been updated [refresh]" bottom bar) so it was possible to update the recipes without having to update the app itself (especially useful to handle services DOM updates that could break `getMessages` implementations). That is something we will want to re-enable further down the road, but I agree with taking-away the archiving step from the contributor for the reasons you mentioned.

vantezzen:
> I think the command should fail (exit code != 0) if one or more recipes couldn't be packaged, what do you think?

Yes, you're right, that sounds like the right idea. This way Travis will also immediately show us if there are problems with the recipes on GitHub

> I also think we can remove the meetfranz- prefix from the LastPass recipe identifier to harmonize

Oh, when was this added in the first place? Yes, we can rename it back

> the archives were useful for recipes "micro-updates"

When we decide to implement this into Ferdi we can implement the packaging step into ferdi-server, too. This way, ferdi-server can always host the newest versions of all recipes.

kytwb (approved):

notNSANE:
tried the whatsapp fix in user.js and didn't work. Maybe something else changed?

PR #301, opened 2020-09-22
Gmail getMessages produces wrong value

stephenpapierski:
When the gmail inbox is set to something like priority inbox, the badge count does not lineup with the number displayed next to "Inbox". Instead it is grabbing the total number of unread messages in the traditional inbox. With priority inbox, you could have thousands of messages that have been filtered from your priority inbox, but they are still showing up in the message count.
image

I modified getMessages to grab the value displayed next to "Inbox" so that it adjusts according to the type of inbox you have selected.
image

### DIFF
--- a/recipes/gmail/package.json
+++ b/recipes/gmail/package.json
@@ -1,7 +1,7 @@
 {
   "id": "gmail",
   "name": "Gmail",
-  "version": "1.3.3",
+  "version": "1.3.4",
   "description": "Gmail",
   "main": "index.js",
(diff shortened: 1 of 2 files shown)


vantezzen:
Thank you for your contribution! Could you please also increase the version number in the recipe's package.json - this way Ferdi will detect changes in the recipe.

https://github.com/getferdi/recipes/blob/2bb318a9ea93a0735949c21b8716f81fbd7ecc37/recipes/gmail/package.json#L4

PR #302, opened 2020-09-23
Update user agent for OWA and Outlook

StormPooper:
Allows enabling Desktop Notifications. I tried the Gmail approach of updating request headers but it didn't work.


kytwb:
@StormPooper thank you!

PR #307, opened 2020-10-02
More careful Gmail unread count detection

kris7t:
10a670fe33bd7fab841ddaf4674deb11477f2267 changed the unread count detection to grab the number besides the Inbox link instead of the total number of unread messages in the traditional inbox. However, this detection may fail in there are no unread messages in the inbox and report the number of unread messages from a different category instead.

This pull request combines the previous and the combined approach by first finding the "Inbox" link as in the previous approach, but it extracts the number of unread messages from the number beside it (instead of the aria-label with the total number of unread messages).

Thus, the number of unread messages in the (priority) inbox is correctly reported even if there are unread messages in other categories.

### DIFF
--- a/recipes/gmail/package.json
+++ b/recipes/gmail/package.json
@@ -1,7 +1,7 @@
 {
   "id": "gmail",
   "name": "Gmail",
-  "version": "1.3.4",
+  "version": "1.3.5",
   "description": "Gmail",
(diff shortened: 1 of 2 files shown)


vantezzen:
Thank you for your contibution! Again, this change will be released with the next update of Ferdi.

PR #308, opened 2020-10-02
Spoof Chrome plugins for Skype

kris7t:
Skype for Web requires either Chrome or Edge, but refuses to cooperate with Chromium. The detection of (non-)Chromium is based on navigator.plugins: https://github.com/Eloston/ungoogled-chromium/issues/1010#issuecomment-643740388

Since we cannot rebuild Chromium to report a different plugin identity as they did it in `ungoogled-chromium`, we instead patch Plugins.prototype to do the spoofing. This is a pure JavaScript approach, but, unlike the JavaScript approach in the issue linked above, it still lets Skype for Web to load.

Closes https://github.com/getferdi/ferdi/issues/973.

### DIFF
--- a/recipes/skype/package.json
+++ b/recipes/skype/package.json
@@ -1,7 +1,7 @@
 {
   "id": "skype",
   "name": "Skype",
-  "version": "2.0.0",
+  "version": "2.1.0",
   "description": "Skype",
   "main": "index.js",
   "author": "Stefan Malzner <stefan@adlk.io>",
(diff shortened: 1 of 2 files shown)


vantezzen:
Thank you for taking the time to debug this problem! Could you please also update the version number in the recipe's "package.json" (in recipes/skype/package.json)?

kris7t:
@vantezzen Sure! I guess the same applies to https://github.com/getferdi/recipes/pull/307?

vantezzen:
Yes, exactly

vantezzen:
Thank you for your contribution! The changes to the Skype recipe will be published with the next update of Ferdi

PR #309, opened 2020-10-03
Update Zoho icon

tofran:
Updates Zoho mail icon.

Before -> After


vantezzen:

PR #310, opened 2020-10-04
Added recipe for yahoo mail.

vraravam:


recipes/yahoo-mail/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "id": "yahooMail",
+  "name": "yahoo-mail",
```
vantezzen, line 3:
```suggestion
  "name": "Yahoo Mail",
```

recipes/yahoo-mail/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "id": "yahooMail",
```
vantezzen, line 2:
```suggestion
  "id": "yahoo-mail",
```

vantezzen (changes requested):
Thank you for taking the time to add your Yahoo Mail recipe to the Ferdi recipe repository.

The recipe looks good, but I have two suggestions for your package.json:
- You should change the "name" in the package.json to "Yahoo Mail" instead of "yahoo-mail". The attribute is the name shown inside Ferdi
- You should change the "id" to "yahoo-mail". The ID should be the same as the name of the folder the recipe is the (i.e. you named the folder `recipes/yahoo-mail`, thus the ID should be `yahoo-mail`)

recipes/yahoo-mail/package.json
vraravam: tx @vantezzen  - changed the 2 fields as per your comments.

recipes/yahoo-mail/package.json
vraravam: Done

vantezzen:
Thank you for your contribution! The recipe will be available in the next version of Ferdi.

PR #312, opened 2020-10-04
Add recipe for Slite

tofran:
Ferdi recipe for [Slite](https://slite.com/)

Tested and working great:

Note: although the `querySelector` is using the `data-test-id` attribute, after closer inspection it's the one that makes more sense and looks less subject to change (the whole app uses keys like this).


tofran:
Woops, I'll fix ASAP

kytwb:
@tofran nice, thank you!

PR #313, opened 2020-10-05
Minify images using ImageOptim (lossless compression)

vraravam:
I found that the images distributed with ferdi are unnecessarily big. Using ImageOptim (which has many lossless image processors/algorithms), I was able to decrease the download size - thus reducing your cost for hosting installation artifacts, as well as reducing the bandwidth requirements for all users of ferdi. Though this has not been raised as an issue, I would very much like if this is taken on the merit and merged.

Total size of the `recipes` folder:
Before compression: ~ 15M
After compression: ~ 11M

(which is almost 30% reduction!!)


vraravam:
These are the options chosen for the software used:
Screen Shot 2020-10-05 at 7 08 23 AM
Screen Shot 2020-10-05 at 7 08 31 AM

kytwb:
@vraravam thank you for your contribution!

Maybe this is something we can automate in the packaging script using [one of these modules](https://www.npmjs.com/search?q=imageoptim)?

vraravam:
thanks @kytwb - someone else had already pointed me to the same - and I am in the process of implementing and raising a PR for that cli-based script.

vraravam:
@kytwb - I have raised a different PR for the above suggestion  #323 . Please approve and merge

kytwb:
@vraravam done, thank you!

PR #325, opened 2020-10-09
Add recipe for Microsoft To Do

arioki1:


kytwb (approved):

kytwb:
@arioki1 thank you!

recipes/mstodo/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "id": "mstodo",
+  "name": "Microsoft To-do",
+  "version": "1.0.0",
+  "description": "Microsoft To-do",
```
kytwb, line 5: Can you rename the service to "Microsoft To Do" with this case please?

kytwb (changes requested):

recipes/mstodo/package.json
arioki1: i've changed that. @kytwb

PR #326, opened 2020-10-10
Minify images using ImageOptim (lossless compression)

vraravam:
Second try. This time around, the script itself reports the size before and after processing. Based on this, the number of files touched is much smaller. I have also manually verified each touched image by opening it and checking that it has some image (and is not a zero-byte file).

@kytwb - please review.


package.json
```diff
@@ -8,7 +8,8 @@
   },
   "scripts": {
     "package": "node scripts/package.js",
-    "create": "node scripts/create.js"
+    "create": "node scripts/create.js",
+    "minify-images": "./minify-images.sh"
```
kytwb, line 12: Should we move this in the `scripts` folder?

vraravam: sure - i can do that @kytwb

vraravam: Done in the amended commit @kytwb

kytwb:
@vraravam thanks for the iteration!

PR #343, opened 2020-10-25
Add recipe for Nextcloud

eandersons:
[Nextcloud](https://nextcloud.com) is a safe home for all your data – community-driven, free & open source.

Should I remove Nextcloud related recipes I added before:

- `nextcloud-calendar`;
- `nextcloud-carnet`;
- `nextcloud-cospend`;
- `nextcloud-tasks`?

These and all other Nextcloud applications can now be accessed from the new Nextcloud service.


kytwb:
@eandersons thank you for following up on the Nextcloud recipes!

Please go ahead and delete the deprecated ones; I don't think it will cause any issues.

vantezzen:
> Please go ahead and delete the deprecated ones

I think it might be good to keep those, wouldn't it? This way, people who don't want to add their Nextcloud as a whole but rather only use specific parts (e.g. the calendar) can do so.

kytwb:
@vantezzen:

> These and all other Nextcloud applications can now be accessed from the new Nextcloud service.

So I assumed @eandersons motive was valid as a user and as the maintainer of the previous recipes.
I don't have

eandersons:
> I think it might be good to keep those, wouldn't it? This way, people who don't want to add their Nextcloud as a whole but rather only use specific parts (e.g. the calendar) can do so.

@vantezzen, good point!

In that case I will try to fix notifications count for existing Nextcloud recipes I added before and add a couple more Ferdi services for collaboration oriented Nextcloud apps.

PR #350, opened 2020-11-04
Fix unread emails css selector

tpapamichail:


vantezzen:
Thank you for your contribution. Please make sure to also update the version number of the recipe as described in https://github.com/getferdi/recipes/blob/master/docs/updating.md#4-updating-the-version-number, otherwise Ferdi won't notive the update

PR #353, opened 2020-11-10
Nextcloud Calendar: Display only calendar related notifications and their count

eandersons:
`service.css` was updated to hide notifications that are not related to Nextcloud Calendar and "Dismiss all notifications" as this action would dismiss hidden notifications as well.
`webview.js` was updated to fix notifications count selector.


vantezzen:

PR #354, opened 2020-11-10
Nextcloud Talk: Hide unrelated notifications and fix notifications count

eandersons:
`service.css` was updated to:

- unhide contacts menu;
- hide notifications that are not related to Nextcloud Talk;
- hide "Dismiss all notifications" as this action would also dismiss hidden
  notifications.

`webview.js` was updated to fix notifications (both direct and indirect) count retrieval.

`README.md` was added.


vantezzen:

PR #356, opened 2020-11-10
Habitica: Add notification count.

iansealy:
`webview.js` was updated to add a notification count.


vantezzen:

PR #359, opened 2020-11-11
Roundcube: Fix unread email count.

iansealy:
`webview.js` was updated to fix updating the unread email count.


vantezzen:

Issue #135, opened 2020-04-22
Google Classroom


opekope2:
Name of the service: Google Classroom
URL of the service: https://classroom.google.com
(I needed to enter https://classroom.google.com/u/0 to show the sign in page)

TanZng:
WIP! (Work In Progress)
Is "ready" but it have some issues while using dark mode.
You can check it out here 👇🏼
https://github.com/TanZng/ferdi-googleclassroom

Nokia808:
Any progression in this issue? It is important to add this service in this black days when global attacks by COVID19 & global shutdown with currently incoming winter.

vantezzen:
This was just added to the repository. The service will be available in the next update of Ferdi

Nokia808:
Too much thanks for that !!

flawnn:
It disappeared for me, lol

PR #363, opened 2020-11-15
Add Google Classroom recipe

TanZng:
Google Classroom with DarkMode. This solves #135


vantezzen:

PR #398, opened 2020-12-31
Synchronize Outlook services and correct message count

cromefire:
Synchronized both outlook services because they are the same more or less.

Also counts the email of all favorites now and not every folder, like before (which counted Inbox twice for example).


kytwb:
@cromefire thank you!

PR #401, opened 2021-01-07
change user-agent to firefox to get around google login issue

arioki1:
I tried to fix this issue by changing user-agent to firefox 
https://github.com/getferdi/ferdi/issues/1117
https://github.com/getferdi/ferdi/issues/894

### DIFF
--- a/recipes/hangouts/index.js
+++ b/recipes/hangouts/index.js
@@ -4,7 +4,7 @@ module.exports = (Franz) =>
       return [
         {
           headers: {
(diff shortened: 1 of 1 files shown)


vraravam:
With the latest changes landing in the main ferdi that allows the user to have more control over the user-agent both globally and per recipe, i think this PR is no longer necessary? IF so, please close it

vraravam:
Closing as there's no word from the OP and since the UA is not editable by the user in the preferences, it makes this PR unnecessary.

PR #429, opened 2021-01-28
Add recipe for BiP

arioki1:


kytwb:
@arioki1 thank you!

erturkvarol:
Hi,
When I tried to load the BiP service, Ferdi can't load the square code and it shows empty space instead of SQ.
Do you have any suggestions to solve this, or it is a bug?
Thanks,

PR #431, opened 2021-02-01
Added recipe for productivity app Clockify

sunnysinghnitb:


kytwb:
@mmso thank you!

laryn:
Was this meant to go in the root folder? Seems like it should be in `/recipes/` so I'm guessing it was accidental!

kytwb:
@laryn good catch, fixed it via 30a759f. thanks!

PR #432, opened 2021-02-01
Fix null element for proton-mail

mmso:
Fix the following error when the element is not displayed

Screenshot 2021-02-01 at 15 13 44


kytwb:
@mmso thank you!

PR #437, opened 2021-02-14
Add recipe for Canvas

Sagir-mo:
Add recipe for Canvas


Sagir-mo:
Users can add their school's canvas without copying the custom URL by simply typing their school's initials as the teamID.

Sagir-mo:
The initial plan was that users would provide the custom URL for their school but this means that they would have to make sure to copy their dashboard's URL and not just the homepage URL. The teamID method simplifies this and has worked for all the schools tested. However, providing the custom URL would ensure that canvas behaves as expected for all users/schools in case some school(s) are a little more customized. Please advice.

kytwb:
@Sagir-mo thank you for your contribution.

Let's go with teamID for now and advise if it becomes an issue to others users.

PR #441, opened 2021-02-25
updated whatsapp unread messages detection again

bpwned:


kytwb:
@bpwned thank you for the pull request, but we are going with @TheBoroer method (see https://github.com/getferdi/recipes/pull/446).

bpwned:
That makes sense.

PR #446, opened 2021-03-04
Fix WhatsApp unread notification badge with new method

TheBoroer:
Fixed web whatsapp notification badges again. 
I decided to fall back on traversing through the DOM Instead of using CSS classes since it seems less prone to future changes.
CSS classes suck because they get generated/changed often ... probably random hash replacements done by their build tool when js/css files are bundled).

Any thoughts or comments are appreciated.


kytwb:
@TheBoroer well thought, thanks!

PR #477, opened 2021-04-26
Fixing issue with non-sso login for gmail + gdrive not working due to incompatible user-agent/browser version

vraravam:
For eg, my company's SSO login though Okta works fine, but my personal gMail using the username/password doesn't.


vraravam:
@kytwb - would you be able to review/approve the recipe PRs as well? TIA

vantezzen (approved):

PR #480, opened 2021-04-29
Minify images

vraravam:
Some PRs/commits had brought in images that needed to be minified so as to minimize the resulting app image.


vraravam:
Screen Shot 2021-04-29 at 8 57 18 PM

Just a simple comparison of the same code without minification (in the `/Applications/` folder vs the one with the minified images (above PR) present in the `./out/mac/` folder.

kytwb:
@vraravam thanks!

PR #500, opened 2021-05-13
Add service for Office 365 GCC Outlook

JoKneeMo:
Adds service for the GCC, GCC High, and GCC DoD versions of Outlook


recipes/office365-gcc/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "Office 365 Outlook Web App for Government Community Cloud",
+  "main": "index.js",
+  "author": "JoKneeMo <https://github.com/JoKneeMo>",
+  "contributors": [
+    "Adam Birds <adam.birds@adbtechltd.co.uk>",
+    "nonAlgebraic <https://github.com/nonAlgebraic>"
+  ],
+  "license": "MIT",
+  "config": {
+    "serviceURL": "https://outlook.office365.us/mail/",
```
vraravam, line 14: Is this the only difference between this recipe and the regular office35-owa? If so, wouldn't it be better to have the ability to specify a custom `serviceURL` using `hasHostedOption` and `hasCustomUrl`?

vraravam (changes requested):

recipes/office365-gcc/package.json
JoKneeMo: Yes, that's the only difference. I honestly considered a change on the existing owa version, but thought it would be easier to not have to type in the URL. I think an ideal solution would be if recipes supported multiple named serviceURLs like below, and then the user can pick their endpoint from a dropdown menu.

Example:

```json
{
  "config": {
    "serviceURL": {
      "Main (USA)": "https://outlook.office365.com/mail/",
      "GCC (USA Gov)": "https://outlook.office365.us/mail/",
      "Germany": "https://outlook.office.de/mail/"
    }
  }
}
```

vraravam: I partially agree with you, but at the same time, the SaaS owners might choose to change DNS names at their will. For a tool like Ferdi, it is better for the end user to be in control of the config metadata (`serviceUrl` in this case). So, as long as we provide them the ability to configure such items from the preferences screen, we can be shielded from having to release new app versions for each such change.

At the same time, what you suggest is a good enhancement - but it will have to be supported by the upstream (Franz). For now, please close this PR - and raise a feature enhancement against the Franz repo.

PR #519, opened 2021-05-21
Mattermost: fix unreads count

kemenaran:
Mattermost's "Unread messages" badge was broken for two reasons.

1. Mattermost introduced a new configurable Sidebar, which broke the selectors used by the recipe.

   _Note that the legacy sidebar can still be enabled, and may also be displayed by older Mattermost instances. So for compatibility, the legacy selectors are kept._
2. Direct group messages (direct messages sent to more than one person) weren't counted at all.

   Now these messages are properly counted as direct messages (because that's how Mattermost UI considers them).

Fixes https://github.com/getferdi/ferdi/issues/1110

### DIFF
--- a/recipes/mattermost/package.json
+++ b/recipes/mattermost/package.json
@@ -1,7 +1,7 @@
 {
   "id": "mattermost",
   "name": "Mattermost",
-  "version": "1.2.2",
+  "version": "1.2.3",
   "description": "Mattermost",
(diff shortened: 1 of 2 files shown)


vraravam (approved):
LGTM

vraravam:
Thanks for the PR submission @kemenaran !

PR #535, opened 2021-05-27
Recipe context isolation

kris7t:
Lets us enable context isolation is recipe webviews.

This very much needs testing, especially the

* fastmail
* googlecalendar
* googledrive

recipes that I didn't have the chance to test. Also, we should look for any other recipes that need to be updated for context isolation.

This must be merged simultaneously with https://github.com/getferdi/ferdi/pull/1456.


vraravam:
Please add the auto-generated `all.json` into the PR?

PR #583, opened 2021-07-05
Remove png icons

vraravam:
While researching [this discussion](https://github.com/getferdi/ferdi/discussions/1601), I chanced upon the fact that Ferdi does not depend on the `*.png` images (maybe it did at an earlier date).
By removing the remaining methods that reference the png image files, we can then safely remove the images themselves in the recipes, thus making the final installer file smaller in size while also removing dead code.


kris7t:
LGTM except for https://github.com/getferdi/ferdi/pull/1606#issuecomment-874356061

PR #585, opened 2021-07-06
fix: Rocket.Chat tab doesn't display notifications

mhatvan:
Closes https://github.com/getferdi/ferdi/issues/1599

### DIFF
--- a/all.json
+++ b/all.json
@@ -1264,7 +1264,7 @@
     "featured": false,
     "id": "rocketchat",
     "name": "Rocket.Chat",
-    "version": "1.0.9",
+    "version": "1.0.10",
     "icons": {
       "svg": "https://cdn.jsdelivr.net/gh/getferdi/recipes/recipes/rocketchat/icon.svg"
     }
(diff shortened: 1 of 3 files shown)


mhatvan:
@vraravam Quite certain that the code is correct but didn't find a way to test it.
How to test the badge appearing? When I `console.log` locally it never seems to go into the `webview.js` file...

vraravam:
@mhatvan - you can do it in 2 ways:
1. Run the app in dev mode, register the service that you want to test, make changes in the `FerdiDev` profile folder (same parent location as the main runtime - just that the folder is `FerdiDev` as opposed to `Ferdi`). You can also use the `View > Toggle Service Developer Tools` menu item to review and check whether the css selectors are fine.
2. Run the app in real (production) mode. Sign up for the service (rocket chat), and close Ferdi, edit the recipe config files and test.

But you will still need to have signed up for rocket.chat and sign into your real account, and have someone send you a message that you then *do not read* - so that the "unread count" stays there.

mhatvan:
Thank you for the detailed explanation!

I used the second way, edited the prod app with my code changes, both direct and indirect messages work as expected ;)

vraravam (approved):
LGTM (I don't have a login into rocket.chat - trusting that @mhatvan has run this locally and verified that this works)

PR #587, opened 2021-07-07
Fixing calendar recipe to enable dark reader

prasann:
**Issue**

Inside calendar recipe, we are creating a modal and replace `window.alert` As part of this, we are attaching a modal to a dom. But due to a race condition, the `document.body` isn't present causing an undefined exception. 

Due to this exception, "Enabling dark mode" isn't working since the recipe isn't fully loaded.

**Fix**
Now, we are waiting for the `document.body` to be present before attaching the modal. This eliminates the error and loads the recipe successfully.

This resolved the calendar issue mentioned here: https://github.com/getferdi/ferdi/issues/1604

### DIFF
--- a/all.json
+++ b/all.json
@@ -84,7 +84,7 @@
     "featured": false,
     "id": "basecamp",
     "name": "basecamp",
-    "version": "1.0.1",
+    "version": "1.0.2",
     "icons": {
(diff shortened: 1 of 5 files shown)


vraravam:
Very nice catch @prasann ! Thanks for the fix!

vraravam (approved):
LGTM

vraravam:
@all-contributors please add @prasann as code contributor

PR #600, opened 2021-07-20
Fix Screen Share not working on Discord

eitchtee:
Adds a polyfill that replaces Discord's non-working native screen share window for a very basic working one.
See https://github.com/electron/electron/issues/16513#issuecomment-602070250 for more information.

---

Screen Share window on a chromium browser
image

Screen share on Discord's own app
image

Replacement Screen Share window introduced by this PR
image
_Despite the design change, the only thing missing is a close button._

Screen share working on Ferdi 5.6
image

_PS.: While this works most of the time, sometimes the shared screen will flicker, restarting the transmission a few times seems to fix this. No idea what might be the cause._

---

Theoretically we could apply this to other problematic services mentioned on https://github.com/getferdi/ferdi/issues/1541, requires testing though.

### DIFF
--- a/recipes/discord/webview.js
+++ b/recipes/discord/webview.js
@@ -29,3 +29,64 @@ module.exports = Franz => {
   Franz.loop(getMessages);
   Franz.injectCSS(_path.default.join(__dirname, 'service.css'));
 };
+
+// Polyfill to enable Screen Sharing
+// See https://github.com/electron/electron/issues/16513#issuecomment-602070250
+const { desktopCapturer } = require('electron')
+window.navigator.mediaDevices.getDisplayMedia = () => {
+  return new Promise(async (resolve, reject) => {
+    try {
+      const sources = await desktopCapturer.getSources({ types: ['screen', 'window'] })
+
+      const selectionElem = document.createElement('div')
+      selectionElem.classList = 'desktop-capturer-selection'
+      selectionElem.innerHTML = `
+        <div class="desktop-capturer-selection__scroller">
+          <ul class="desktop-capturer-selection__list">
+            ${sources.map(({id, name, thumbnail, display_id, appIcon}) => `
+              <li class="desktop-capturer-selection__item">
+                <button class="desktop-capturer-selection__btn" data-id="${id}" title="${name}">
+                  <img class="desktop-capturer-selection__thumbnail" src="${thumbnail.toDataURL()}" />
+                  <span class="desktop-capturer-selection__name">${name}</span>
+                </button>
+              </li>
+            `).join('')}
+          </ul>
+        </div>
+      `
+      document.body.appendChild(selectionElem)
+
+      document.querySelectorAll('.desktop-capturer-selection__btn')
+        .forEach(button => {
+          button.addEventListener('click', async () => {
+            try {
+              const id = button.getAttribute('data-id')
+              const source = sources.find(source => source.id === id)
+              if(!source) {
+                throw new Error(`Source with id ${id} does not exist`)
+              }
+              
+              const stream = await window.navigator.mediaDevices.getUserMedia({
+                audio: false,
+                video: {
+                  mandatory: {
+                    chromeMediaSource: 'desktop',
+                    chromeMediaSourceId: source.id
+                  }
+                }
+              })
+              resolve(stream)
+
+              selectionElem.remove()
+            } catch (err) {
+              console.error('Error selecting desktop capture source:', err)
+              reject(err)
+            }
+          })
+        })
+    } catch (err) {
+      console.error('Error displaying desktop capture sources:', err)
+      reject(err)
+    }
+  })
+}
--- a/recipes/discord/webview.js
+++ b/recipes/discord/webview.js
@@ -33,7 +33,7 @@ module.exports = Franz => {
 // Polyfill to enable Screen Sharing
 // See https://github.com/electron/electron/issues/16513#issuecomment-602070250
 const { desktopCapturer } = require('electron')
-window.navigator.mediaDevices.getDisplayMedia = () => {
+navigator.mediaDevices.getDisplayMedia = () => {
   return new Promise(async (resolve, reject) => {
     try {
       const sources = await desktopCapturer.getSources({ types: ['screen', 'window'] })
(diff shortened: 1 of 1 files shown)


vraravam:
> Theoretically we could apply this to other problematic services mentioned on getferdi/ferdi#1541, requires testing though.

Very nice work @eitchtee - yes, I agree this needs some testing. But, if we can get one more corroboration from another Ferdi user that this works, I can merge this PR. (Unfortunately, I dont have a discord signup to test this out).

eitchtee:
Thank you @vraravam 

Fingers crossed somebody else pick this up for testing 🤞🏻. Unfortunately I also don't use the other services mentioned, so unable to help there.

I'm already using these changes on my computer, so I will keep this updated in case I run into any problems. Also might have a look into styling the window and the flickering issue if I have some spare time on the weekend.

Keep up the good work.

recipes/discord/webview.js
```diff
@@ -29,3 +29,64 @@ module.exports = Franz => {
   Franz.loop(getMessages);
   Franz.injectCSS(_path.default.join(__dirname, 'service.css'));
 };
+
+// Polyfill to enable Screen Sharing
```
vraravam, line 33: Is it possible to move this into its own file and inject it in here?

JesseBellingham:
@eitchtee @vraravam I can confirm that this fix works. I was able to screenshare with a friend and they confirmed that they could see my screen.

Version: 5.6.0
Electron: 13.1.7
Chrome: 91.0.4472.124
Node.js: 14.16.0
Platform: macOS Big Sur
Arch: x64
Build date: Mon Jul 19 2021 01:17:34 GMT+1000 (Australian Eastern Standard Time)
Git SHA: 7b09fe7
Git branch: release

eitchtee:
@jbellingham Excellent! Glad to see it also works on MacOS. I've just pushed a new commit with some styling and a crude close button on the top right, mind giving some feedback?

image

recipes/discord/webview.js
eitchtee: Sounds possible, not sure about the technicalities, will have a look at it later.

An idea I had would be to have a config tag on `package.json`, which when set to true would preload this file on the WebView, like what already happens with `webview.js` itself. This would allow for some freedom while not disrupting every service, as it could be set individually on each recipe.

Also, I was doing some research and found the [exact same code on Ferdi codebase](https://github.com/getferdi/ferdi/blob/33123c354b79f7951423dd75097b11e7eb075f99/src/webview/screenshare.js), know what that might be about?

vraravam: > An idea I had would be to have a config tag on package.json, which when set to true would preload this file on the WebView, like what already happens with webview.js itself. This would allow for some freedom while not disrupting every service, as it could be set individually on each recipe.

Exactly my thoughts on the reason for my request. For ease of understanding, we can call it `hasScreenSharing`.

Oh - I didnt realize that this code already existed. Let me spend some time in understanding the history.

vraravam: fyi: this file came into the codebase *very recently* from (this comment](https://github.com/getferdi/ferdi/pull/1456#discussion_r641194876).
Maybe @kris7t can shed some light on this

vraravam: @jbellingham - can I request you to use the contents of the [file](https://github.com/getferdi/ferdi/blob/33123c354b79f7951423dd75097b11e7eb075f99/src/webview/screenshare.js) to check if the polyfill works? That way, the work that @kris7t did in the earlier PR can also be validated and we dont end up with duplicate code.

kris7t:
Wait, don't we have a global screen sharing polyfill (which got a close button in https://github.com/getferdi/ferdi/pull/1460)? Why does Discord need one on its own?

Also note that screen sharing from recipes (as opposed to the global one, which is injected by Ferdi itself) would probably stop working after https://github.com/getferdi/ferdi/pull/1456

Are you sure you're running the latest version of Ferdi, which includes the global screen sharing code? If so, we should try and debug the Discord recipe and see why global screen sharing is broken there, instead of duplicating the code.

recipes/discord/webview.js
vraravam: Lets continue this conversation in the main [thread](https://github.com/getferdi/recipes/pull/600#issuecomment-883404141)

vraravam:
> we should try and debug the Discord recipe and see why global screen sharing is broken there, instead of duplicating the code.

If we are able to solve this - i think it can warrant a 5.6.1 release all by itself - since it can unlock (imo, but 🤞) for all other screensharing services.

kris7t:
Btw, for me, screen sharing in MS Teams works in the current release, so I'm not sure what's broken with Discord.

eitchtee:
> Wait, don't we have a global screen sharing polyfill (which got a close button in [getferdi/ferdi#1460](https://github.com/getferdi/ferdi/pull/1460), but was around since much earlier)? Why does Discord need one on its own?
> 
> Also note that screen sharing from recipes (as opposed to the global one, which is injected by Ferdi itself) would probably stop working after [getferdi/ferdi#1456](https://github.com/getferdi/ferdi/pull/1456)
> 
> Are you sure you're running the latest version of Ferdi, which includes the global screen sharing code? If so, we should try and debug the Discord recipe and see why global screen sharing is broken there, instead of duplicating the code.

I discovered the Polyfill today on Ferdi's code, didn't know it had already been implemented, but yes, screen sharing wasn't working on Discord on the latest version. I've also tried deleting the service and recreating it, since the problematic one was created on an older version, but it still doesn't work and the problem seems to extend to other services as per https://github.com/getferdi/ferdi/issues/1541

image

s2XahPC8ZD
_Clicking on the Screen Share button doesn't work._

kris7t:
Can you try whether applying https://github.com/getferdi/ferdi/pull/1456#event-5035976832 helps with the global polyfill?

eitchtee:
> Can you try whether applying [getferdi/ferdi#1456 (comment)](https://github.com/getferdi/ferdi/pull/1456#event-5035976832) helps with the global polyfill?

By try you mean building from source? Or is there an already built version somewhere?
I can do that sure, but gonna take a while since I don't have the environment setup at the moment.

vraravam:
@eitchtee - I think what he meant was to copy/paste the contents of the file from Ferdi that you found (https://github.com/getferdi/ferdi/blob/33123c354b79f7951423dd75097b11e7eb075f99/src/webview/screenshare.js) - and use it within the recipe's webview.js instead of the content from this PR. Please try and check if that works. imo, it should not need to be built from source.

kris7t:
Not really, I meant running with the contextIsolation patch. That code from `webview.js` won't work at all without all the other supporting code from the contextIsolation patch that exposes our wrapper of the screensharing API from the isolated world to the main world.

eitchtee:
Sorry, I couldn't get the environment working, was going to try again tomorrow as work got in the way during the week.

But as I said on the original issue, screen sharing seems to be functional now without the need to change `webview.js`, thus making this RP unnecessary. Feel free to close this.

image

Thanks for the amazing work @vraravam and @kris7t, hope to contribute more soon.

vraravam:
Closing this as per the last comment. Thanks for your contribution @eitchtee !

PR #603, opened 2021-08-08
build: migrate from npm to pnpm

mhatvan:
- remove package-lock.json
- add pnpm-lock.yaml

Precondition to finish PR: https://github.com/getferdi/ferdi/pull/1714


scripts/create.js
```diff
@@ -1,64 +1,70 @@
 /**
  * Create a new recipe for your service
  */
-const fs = require('fs-extra');
-const path = require('path');
-const open = require('open');
+const fs = require("fs-extra");
```
vraravam, line 4: Didn't we decide to use single quotes all over?

vraravam (changes requested):

scripts/create.js
mhatvan: You are correct, this is the default setting of prettier.

We don't have a `.prettierrc.js` config in the recipes project, therefore this happened.

I will add one.

mhatvan: Done.

vraravam (approved):

PR #604, opened 2021-08-08
feat: add google chat

mhatvan:
Closes https://github.com/getferdi/ferdi/issues/1615

Currently, Google Hangouts and Google Chat seem to be two different services. 

There seems to be no redirect happening from old Hangouts to new Google Chat. 

That's why I added Google Chat, so users can choose to their liking.


mhatvan:
## Still to do

- fix unsupported browser issue in Google Chat
- count notifications with webview.js

vraravam:
The `hangoutschat` is the same recipe. We should not create a duplicate just for the sake of the different name. The other service has the same `serviceUrl`. Please close this PR.

vraravam:
We need to either implement the solution for the aliases/tags as discussed [here](https://github.com/getferdi/ferdi/issues/1614#issuecomment-875211630). Or a simpler change would be to include this new name in the current existing recipe - thus making it searchable.

mhatvan:
Ohhh, that is highly confusing with the duplicated services..
Just saw hangouts, not hangoutschat

vraravam:
Please close this PR or revert these new files and only have the name-change in the pre-existing `package.json` file.

vraravam:
Closing since this is now handled by the aliases PR #607

PR #613, opened 2021-08-24
Add Google News recipe

niutech:


vraravam (changes requested):
Thanks for the contribution @niutech . Could I please request that you remove the empty file `webview.js` - since these are no longer required.

vraravam (approved):

PR #673, opened 2021-09-04
Changed WhatsApp recipe unread count logic to get value from document title 

dolukhanov:
Changed unread message logic to get value from document title. This gives a more accurate count, including messages that are not currently in view.


recipes/whatsapp/webview.js
```diff
@@ -26,26 +26,14 @@ window.addEventListener('beforeunload', async () => {
...
-      }
+    if (document.title != null && document.title.length > 0 && document.title.indexOf("(") !== -1 && document.title.indexOf(")") !== -1)
+    {
+      let titleCount;
+      titleCount = document.title.substring(document.title.indexOf("(") + 1, document.title.lastIndexOf(")"));
+      count = parseInt(titleCount);
     }
 
-    Ferdi.setBadge(count, indirectCount);
+    Franz.setBadge(count, count);
```
vraravam, line 36: We have renamed the object to be `Ferdi` instead of `Franz` - please change this line

vraravam (changes requested):
By doing this, we are losing the distinction between direct and indirect counts. Are we sure that this is the direction to go?

vraravam:
This is superceded by another fix that I had made yesterday. Closing

PR #675, opened 2021-09-08
Expose 'known certificate hosts' in ms teams and outlook to try to fix certificate expiry issue

vraravam:
This will not verify that the certificate is valid - it only verifies the hostname is in a known list.


recipes/msteams/webview.js
```diff
@@ -2,6 +2,45 @@ const _path = _interopRequireDefault(require('path'));
...
+      if (knownHosts.find(item => item.indexOf(hostname) !== -1).length > 0) {
+        callback(0);
+      } else {
+        callback(-2);
+      }
+    });
+
+    session.flushStorageData();
+    session.clearStorageData({
+      storages: ['appcache', 'serviceworkers', 'cachestorage', 'websql', 'indexdb'],
```
kris7t, line 32: Is it a good idea to clear the cache on every page load? The costs on repopulating the cache after every page load seems extreme, especially w.r.t bandwidth (which might be an issue on metered connections).

On the other hand, I get that we don't really have a good options to fix this otherwise, short of experimenting with what exactly needs to be cleared and what doesn't.

vraravam: Yes - I got most of the code (about the clearing of cache) from the `google-meet` recipe. Since both gMeet and msTeams have screen-sharing, etc - i thought the storage, service workers, etc would be very similar. Yes - its a trial-and-error going on here unfortunately.

recipes/msteams/webview.js
```diff
@@ -2,6 +2,45 @@ const _path = _interopRequireDefault(require('path'));
 
 function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
 
+window.addEventListener('beforeunload', async () => {
+  try {
+    const { session } = Ferdi.getCurrentWebContents();
+
+    // TODO: Not sure if this 'setCertificateVerifyProc' needs to be called in the 'beforeunload' event or when the app window is loading up
```
vraravam, line 9: @kris7t @mhatvan @sad270 - This TODO comment is something that I am not too sure about.

kris7t: Based on the relevant electron API: https://www.electronjs.org/docs/api/session it looks like this should be called from _outside_ the sandbox. Although maybe we can get away with setting it via the remote module, there doesn't seem to be a need for that, and we needlessly add the overhead of serializing the callback via the remote module (not to mention that we'll run into troubles when we remove the remote module from the recipe webviews).

I'd either

* add a new IPC message to set the verify proc from the _main_ thread, then send the IPC message from the isolated world of the recipe webiew in `recipe.js`; or
* keep relying on the remote module, but do this from the Ferdi render thread (i.e., the same way as for the User-Agent string). A benefit of the latter approach that we could just stuff quirks like these into the `UserAgentModel`.

On the other hand, overriding certificate validation runs the risk of MITM attacks, so a more sophisticated approach might be needed to detect if Teams is sending a genuine certificate that is somehow failing electron's validation.

recipes/msteams/webview.js
vraravam: I'm happy to announce that i remained logged-in into MSTeams via okta even though I closed my laptop (sleep/hibernate) for an hour this afternoon. I will try quitting and restarting Ferdi - and check if i remain logged-in. If that is the case, final test will be to reboot the OS. Once these 3 are working as expected, I can remove this above TODO, and imo, the PR would be ready for merge?

recipes/msteams/webview.js
vraravam: Is there an example of sending an ipc message from a recipe back to the main codebase? I am thinking that the data of the certificate domains could be something that the recipe exposes, while the rest of the code could be in the main codebase triggered via that ipc message. am i thinking in the right direction?

kris7t: Here is an example on how to handle a message from the recipe webview in the main pricess https://github.com/getferdi/ferdi/blob/1eb80721fd30e50369e64d87f84be34dae6c68bd/src/index.js#L500

Of course, here we'd need to get the `session` object from the `webContents`, and set the callback there.

The problem is that when the recipe webview can send a (synchronous/asynchronous) message, it's already too late, because the service webpage is already being loaded. So we might need to handle setting the callbak from the Ferdi rendered process (i.e., _outside_ the recipe webview), similarly to how the `UserAgentModel` works.

vraravam: if by "similar to the UserAgentModel" part - you meant that we can expose a user-editable text area where they can add csv-delimited domains for the recipes? if so, maybe that's the better way than this PR

kris7t: No, I mean run the code that attaches the callback from the same place -- the timing requirements (run before any webpage is loaded) seem fairly similary.

recipes/msteams/webview.js
vraravam: This has been refactored a bit to reuse the `sessionHandler` methods exposed from the main repo.

recipes/msteams/webview.js
vraravam: I think a better style would be related to how `modifyRequestHeaders` is passed from a recipe's `index.js` and into the recipe while its still being **initialized**. Would you agree?
https://github.com/getferdi/ferdi/blob/1eb80721fd30e50369e64d87f84be34dae6c68bd/src/models/Service.js#L250-L253

https://github.com/getferdi/ferdi/blob/1eb80721fd30e50369e64d87f84be34dae6c68bd/src/index.js#L471

vraravam:
Please see https://github.com/getferdi/ferdi/pull/1890 for the corresponding changes in the main repo.

recipes/msteams/webview.js
kris7t: Yeah, that approach looks pretty nice!

recipes/msteams/webview.js
```diff
@@ -2,6 +2,11 @@ const _path = _interopRequireDefault(require('path'));
 
 function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
 
+window.addEventListener('beforeunload', async () => {
```
mhatvan, line 5: Add an explanatory comment for other devs why this is necessary in ms teams.

mhatvan (approved):

recipes/msteams/webview.js
vraravam: done - please review.

PR #680, opened 2021-09-11
update WhatsApp webview.js for non-English locales

bpwned:
The current implementation tries to get the message list parentChatElem through an element with the aria-label "Chat list", which is not named that in other locales. As a replacement, I'm finding all divs with an aria-label attribute and then I choose the tallest one, which should usually be the chat list.


vraravam:
Nice catch @bpwned .

But, for the Pr to be accepted, please bump the version in `package.json` for this change. Also, you will need to run the `pnpm run package` command and add the changed `all.json` into this PR.

bpwned:
Ok, I tried, but something isn't right. Can you help me out? I'm not really a node developer so I'm not sure what's not in order here.

vraravam (approved):
This is good @bpwned - lgtm. 
Thanks for the contribution!

kytwb:
@all-contributors please add @bpwned for code

PR #682, opened 2021-09-11
Added recipe for blackboard and overleaf

cereum:


recipes/protoncalendar/README.md
```diff
@@ -0,0 +1,2 @@
+# ProtonCalendar for Ferdi
+This is a Ferdi recipe for ProtonCalendar
```
vraravam, line 2: proton-calendar is already present - please remove this whole folder from your PR

all.json
```diff
@@ -1070,6 +1079,15 @@
       "svg": "https://cdn.jsdelivr.net/gh/getferdi/recipes/recipes/proton-mail/icon.svg"
     }
   },
+  {
+    "featured": false,
+    "id": "protoncalendar",
+    "name": "ProtonCalendar",
```
vraravam, line 1085: Why are you including proton-calendar - its already present

all.json
```diff
@@ -107,6 +107,15 @@
       "svg": "https://cdn.jsdelivr.net/gh/getferdi/recipes/recipes/bitbucket/icon.svg"
     }
   },
+  {
+    "featured": false,
+    "id": "blackboard",
+    "name": "Blackboard",
+    "version": "1.0.1",
```
vraravam, line 114: For a brand new service-recipe, we would like to start the version with `1.0.0`. Please correct in the corresponding `package.json` and re-generate this file.

recipes/blackboard/webview.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+var _path = _interopRequireDefault(require("path"));
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
+module.exports = Ferdi => {
+  const getMessages = function getMessages() {
+    // TODO: Insert your notification-finding code here
```
vraravam, line 9: This seems like only the recipe that gets generated by default from the template - why do we want to include this new recipe into Ferdi if its nothing more than the shell?

recipes/overleaf/webview.js
```diff
@@ -0,0 +1,15 @@
+"use strict";
+
+var _path = _interopRequireDefault(require("path"));
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
+module.exports = Ferdi => {
+  const getMessages = function getMessages() {
+    // TODO: Insert your notification-finding code here
```
vraravam, line 9: same case here: If its the same as the default-generated shell of the recipe, why do we want to include it into Ferdi?

vraravam (changes requested):

PR #708, opened 2021-09-12
Standardize recipeId and folderName

vraravam:
The `recipeId` should be considered immutable - since the recipe might have been used by a user to configure on their local machine. On the local machine, the `recipeId` is used to create the folder when a user adds a new service.

Added a check to verify that the folder name of the recipe matches the recipeId; Use the recipeId as master to rename internal folder names.


scripts/package.js
```diff
@@ -102,6 +102,9 @@ const compress = (src, dest) => new Promise((resolve, reject) => {
     } else if (!/^[a-zA-Z0-9._\-]+$/.test(config.id)) {
       configErrors.push("The recipe's package.json defines an invalid recipe ID. Please make sure the 'id' field only contains lowercase letters (a-z), numbers (0-9), hyphens (-), periods (.), and underscores (_)");
     }
+    if (config.id !== recipe) {
+      configErrors.push(`The recipe's id (${config.id}) does not match the folder name (${recipe})`);
```
mhatvan, line 106: Perfect, this will help!

mhatvan (approved):

PR #711, opened 2021-09-15
🚀 Add Pipefy Recipe

TanZng:
Add Pipefy Recipe 😃
pipefy


recipes/pipefy/index.js
```diff
@@ -0,0 +1,6 @@
+module.exports = (Ferdi) =>
+  class Pipefy extends Ferdi {
+    overrideUserAgent() {
+      return "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:72.0) Gecko/20100101 Firefox/72.0";
```
vraravam, line 4: This class can be empty - we have moved away from hardcoding the useragent since that is not forwards compatible.

recipes/pipefy/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "pipefy",
+  "name": "Pipefy",
+  "version": "1.0.1",
```
vraravam, line 4: The version here seems to not match the one in the generated file `all.json`. Also, when a recipe first gets inducted, it should start off with `1.0.0`

recipes/pipefy/webview.js
```diff
@@ -0,0 +1,16 @@
+var _path = _interopRequireDefault(require("path"));
+
+function _interopRequireDefault(obj) {
+  return obj && obj.__esModule ? obj : { default: obj };
+}
+
+module.exports = (Ferdi) => {
+  const getMessages = function getMessages() {
+    const FerdiData = document.querySelector("#notifications_button");
```
vraravam, line 9: please use camelCase for variable names.
Also change the variable name from `FerdiData` to something that is more meaningful.

vraravam (changes requested):

package-lock.json
```diff
@@ -0,0 +1,1491 @@
+{
```
vraravam, line 1: Why are we introducing this file?

vraravam (changes requested):

recipes/pipefy/index.js
TanZng: I override the UserAgent to make the google login work correct

recipes/pipefy/package.json
TanZng: version updated 👍🏻

recipes/pipefy/webview.js
TanZng: name changed

package-lock.json
TanZng: oops, file removed

all.json
```diff
@@ -1166,6 +1166,15 @@
       "svg": "https://cdn.jsdelivr.net/gh/getferdi/recipes/recipes/pinterest/icon.svg"
     }
   },
+  {
+    "featured": false,
+    "id": "pipefy",
+    "name": "Pipefy",
+    "version": "1.0.2",
```
vraravam, line 1173: since we are introducing this recipe afresh - please start off with `1.0.0`

recipes/pipefy/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "id": "pipefy",
+  "name": "Pipefy",
+  "version": "1.0.2",
```
vraravam, line 4: use `1.0.0`

vraravam (changes requested):

vraravam:
When developing, you would need to bump up the version so that your local ferdi can recognize the new changes. Once everything works, please change the version back to `1.0.0` before the final PR can be accepted & merged.

all.json
TanZng: version 1.0.0 added

recipes/pipefy/package.json
TanZng: version 1.0.0 added 😄

vraravam (approved):

PR #713, opened 2021-09-18
fix: privileged API access

kris7t:
Privileged browser APIs should be accessed from `webview.js` via the Ferdi parameter to the exported function.

If any service is broken (e.g., screen sharing), then most likely we have to fix `recipe.js` or `screenshare.js`

### DIFF
--- a/recipes/discord/webview.js
+++ b/recipes/discord/webview.js
@@ -1,64 +1,6 @@
-// TODO: Some/most of this is already present in https://github.com/getferdi/ferdi/blob/develop/src/webview/screenshare.js#L5
-
 const _path = _interopRequireDefault(require('path'));
 
 function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
-
-window.navigator.mediaDevices.getDisplayMedia = () => {
-  return new Promise(async (resolve, reject) => {
-    try {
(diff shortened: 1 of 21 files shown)


recipes/discord/webview.js
```diff
@@ -1,65 +1,7 @@
-// TODO: Some/most of this is already present in https://github.com/getferdi/ferdi/blob/develop/src/webview/screenshare.js#L5
```
vraravam, line 1: I think this was reverted from an older version since a user raised a ticket that this didn't work. Can we keep these changes out till we test it please?

vraravam (approved):
This might even fix the issue with the team icon not working in slack. lgtm (except for Discord)

recipes/discord/webview.js
kris7t: I removed this code, because it does precisely _nothing_. I mean, nothing that we overwrite in `window.navigator.mediaDevices` in `webview.js` will have _any_ effect on what the actual discord js code sees. This is context isolation in action.

If the `screenshare.js` code doesn't work with discord (guess we need someone who uses discord to test), then we'll need to create a `webview-unsafe.js` fix that _works with what is already exposed by `screenshare.js`_. But this code can't be it, because it doesn't have the right privileges by definition.

kris7t: Btw, I just tried it (on a discord account I rarely use), and I can join a voice channel and share my screen there without any extra code.

PR #719, opened 2021-09-21
Added trello darkmode

cereum:
Theme from: https://github.com/Xaymar/Trello-Digital-Dark
Added some header button selectors to Layer 1 and Layer 2 to Improve the look.


vraravam (changes requested):
Please bump up the `version` in `package.json` for trello, and then run `pnpm run package`. This should trigger changes in `all.json`. Please include both these changes in the PR before it can be approved.

all.json
```diff
@@ -1638,7 +1638,7 @@
     "featured": false,
     "id": "trello",
     "name": "Trello",
-    "version": "1.1.1",
+    "version": "1.2.1",
```
vraravam, line 1641: Since you are bumping the minor segment, please reset the tiny segment to '0'. ie this should become `1.2.0`

recipes/trello/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "trello",
   "name": "Trello",
-  "version": "1.1.1",
+  "version": "1.2.1",
```
vraravam, line 4: Since you are bumping the minor segment, please reset the tiny segment to '0'. ie this should become `1.2.0`

vraravam (changes requested):

vraravam:
Thanks for the contribution @cereum 🎉🎉🎉🎉

PR #726, opened 2021-09-25
Remove references to 'Ferdi.clearStorageData' and directly send message to the host via ipc

vraravam:
Needs to be merged in tandem with getferdi/ferdi#1987


kris7t:
Wouldn't exposing a helper method on the `Ferdi` object (i.e., `RecipeWebview`) that sends the IPC message instead of sending it manually be cleaner?

vraravam:
> Wouldn't exposing a helper method on the `Ferdi` object (i.e., `RecipeWebview`) that sends the IPC message instead of sending it manually be cleaner?

It was like that before - but, I thought the level of indirection might be unnecessary. If that indirection is actually better, then this PR can be abandoned, and the changes in the main one can be reduced. Please confirm which direction to go for.

vraravam:
@kris7t - I have reverted parts of both the PRs to follow the suggestion that you had. Please review

mhatvan (approved):

kris7t:
> It was like that before - but, I thought the level of indirection might be unnecessary.

In my opinion, adding the indirection is good, because it allows for refactoring the IPC message in the future without having to change the recipes yet again. (I guess the same would apply to the `avatar` and `new-window` IPC messages.)

vraravam:
I agree @kris7t - for the same exact reason as you mention above. I will make the change for the `avatar` and `new-window` as well.

PR #730, opened 2021-10-03
Added hyphens to delimit words

Bluetoothworks:
getferdi/sponsors-only#1712 
Hi there! I have arranged all filenames and IDs according to the convention, changed all camel casings and non-delimiters to delimiters. I hope my changes can be merged.


vraravam:
Thanks for the PR. Though I am 100% behind the naming conventions, and standardizing all file/folder-names to follow this convention, there are some breaking change....

Unfortunately, this will cause a lot of broken services for pre-existing users. Please ensure that a pre-existing ferdi profile (before your changes) can continue to work *after your changes*.

Maybe you can make use of the `aliases` node to setup the old name for searchability. But, how will this solve the recipe being unzipped on the local machine? Please provide a solution for that before this PR can be accepted.

Finally, the icons url in jsdelivr will also need to change to reflect the new names - this also needs to be solved before this PR can be reviewed & merged.

vraravam (changes requested):
Please see comments

PR #756, opened 2021-10-28
fix: darkmode for various recipes

mhatvan:
- inject darkmode properly for various recipes

I wanted to see why darkmode doesn't work for google calendar and then found out that a lot of services have a `darkmode.css`, but its simply not injected by Ferdi.

## Looks good with `darkmode.css`
- googlecalendar
- messenger
- whatsapp

## Not good looking therefore commented out
- linkedin
- slack

## Need someone to take a look if works since I don't use service
- gmail
- jira
- office365
- googleclassroom
- standardnotes
- workplace
- trello
- pipefy


vraravam:
@mhatvan - can you please resolve the conflicts and the force-inject of `darkmode.css`?

mhatvan:
Well, I still have zero clue how they get triggered without force-inject it. Will investigate at some point.

PR #757, opened 2021-10-31
Add Moodle recipe

### LINKED ISSUE #425: Moodle
Name of the service (e.g. WhatsApp): mebis
URL of the service (e.g. whatsapp.com):https://www.mebis.bayern.de/

GitHub URL to a recipe : https://github.com/...

Features you want implemented (optional):

-

vantezzen:
This PR will add a universal recipe for [Moodle](https://moodle.org/), a learning platform used in many schools and universities.

The recipe includes support for unread messages count for the Standard Moodle Theme.

Closes #425


recipes/moodle/webview.js
```diff
@@ -0,0 +1,18 @@
+var _path = _interopRequireDefault(require('path'));
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
+module.exports = Ferdi => {
+  const getMessages = () => {
+    const elements = [...document.querySelectorAll('[data-region="count-container"]')];
```
mhatvan, line 7: `document.querySelectorAll` already returns an array -> pointless spread and cast to array around it. Would also prefer naming it `directCount` instead of just "messages" to be more explicit.

vantezzen: > document.querySelectorAll already returns an array

`document.querySelectorAll` returns a `NodeList`, not an Array (https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll#return_value).

[NodeLists](https://developer.mozilla.org/en-US/docs/Web/API/NodeList) are not the same as arrays, so `.reduce` does not exist on a NodeList:

Screenshot 2021-10-31 at 16 10 13

Using a spread to convert it into an array is part of [the number one answer on StackOverflow on this topic](https://stackoverflow.com/a/3199627/10590162) (apart from `forEach` but as using `reduce` is the simplest option here, I convert to an array).

mhatvan: Ah, that's why you convert it, due to `NodeList`. An alternative would be to for loop over it, which would be a cheaper operation than spread + reduce, but let's just keep it how it is.

mhatvan (approved):

vantezzen:
Oh no, it looks like that merge didn't squash correctly. I'll keep more attention on that in the next PR!

mhatvan:
I updated the repo settings to only allow squash merges.

PR #767, opened 2021-11-20
Toggl recipe: Display 'unread' badge when a timer is running

n-st:
This PR adds an "unread" badge to the Toggl service whenever a timer is running.

It uses Toggl's own JavaScript interface — this lets us get away without searching for hardcoded CSS classes in the DOM, but requires passing information between the service context (where the `toggl` object is available) and the Electron context (where the `ferdi` object is available).
Currently, the only way I've found to do that is with `Ferdi.injectJSUnsafe` — and since the `ferdi` object inside the unsafe JS doesn't offer the usual `loop()` function, I have to reinject my JS on every loop…
Is there a more efficient way to do this?

Side note: The Toggl JS interface also exposes the name of the currently running timer, so we could `ferdi.setDialogTitle`, but that doesn't seem to be available in my Ferdi 5.6.1, either…

### DIFF
--- a/recipes/toggl/webview.js
+++ b/recipes/toggl/webview.js
@@ -1,1 +1,8 @@
-module.exports = Ferdi => Ferdi;
+module.exports = Ferdi => {
+    const path = require('path');
+    const updateBadge = function updateBadge() {
+        Ferdi.injectJSUnsafe(path.join(__dirname, 'webview-unsafe.js'));
+    };
+
+    Ferdi.loop(updateBadge);
+};
--- a/recipes/toggl/webview.js
+++ b/recipes/toggl/webview.js
@@ -1,7 +1,10 @@
+const _path = _interopRequireDefault(require('path'));
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
 module.exports = Ferdi => {
-    const path = require('path');
     const updateBadge = function updateBadge() {
-        Ferdi.injectJSUnsafe(path.join(__dirname, 'webview-unsafe.js'));
+        Ferdi.injectJSUnsafe(_path.default.join(__dirname, 'webview-unsafe.js'));
     };
 
     Ferdi.loop(updateBadge);
--- a/recipes/toggl/package.json
+++ b/recipes/toggl/package.json
@@ -1,7 +1,7 @@
 {
   "id": "toggl",
   "name": "toggl",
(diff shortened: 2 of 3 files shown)


recipes/toggl/webview.js
```diff
@@ -1 +1,8 @@
-module.exports = Ferdi => Ferdi;
+module.exports = Ferdi => {
+    const path = require('path');
```
vraravam, line 2: Please use the pattern found in other recipes where the `webview-unsafe.js` is injected.
Look for `_interopRequireDefault` for eg in the `googlecalendar` recipe

vraravam (changes requested):

recipes/toggl/webview.js
n-st: Done, thank you for the feedback.

vraravam (approved):
lgtm

vraravam:
btw - ferdi is at v5.6.3 right now. You might want to upgrade.

n-st:
@vraravam Does it impact performance or memory usage to inject the JS file so often? Is there a more efficient way (since I'm only doing it so I can use ferdi.loop())?

vraravam:
Not sure - lets wait and watch. Please try out the next nightly and keep an eye out on this factor?

simonszu:
This commit is breaking the linter which is triggered in the pre-commit hook: 

```
simonszu@maenas:~/recipes/recipes$ git commit -m "Added recipe for circuit"

> ferdi-recipes@1.0.0 lint /home/simonszu/recipes
> eslint . --fix


/home/simonszu/recipes/recipes/toggl/webview-unsafe.js
  2:5   error  'ferdi' is not defined  no-undef
  2:27  error  'ferdi' is not defined  no-undef
  7:9   error  'ferdi' is not defined  no-undef
  9:9   error  'ferdi' is not defined  no-undef

✖ 4 problems (4 errors, 0 warnings)

 ELIFECYCLE  Command failed with exit code 1.
husky - pre-commit hook exited with code 1 (error)
```

n-st:
@simonszu Would it satisfy the linter to use `window.ferdi` there?

simonszu:
I think this was already fixed with 
https://github.com/getferdi/recipes/commit/b88f3a009471711299a5878fdadf6d197b505f27

kytwb:
@all-contributors please add @n-st for code

Issue #763, opened 2021-11-08
Skype notifications counter not working


karolkpl:
I have group notification with bunch of unread messages - I can see red badge under group name and "chats" tab inside Skype, but Ferdi shows no badge.

mmalisz:
Hi @vraravam,
I've identified the issue.

The problem lies with the querySelector here: https://github.com/getferdi/recipes/blob/master/recipes/skype/webview.js#L8
This assumes that the user uses english in the recipe.

For example, in polish version the query should be `Czaty` - replacing it in the recipe seems to make the badges work.

However, it's more of a workaround than a real fix.

vraravam:
Unfortunately, I dont use skype. We will appreciate one of the other users to contribute a fix that can work across languages

PR #775, opened 2021-11-25
Fix for #763

ZwS:
Chats tab is actually first in list so with this selector it would find proper element regardless of the lanuguage of service

### DIFF
--- a/all.json
+++ b/all.json
@@ -1467,7 +1467,7 @@
     "featured": false,
     "id": "skype",
     "name": "Skype",
-    "version": "3.2.2",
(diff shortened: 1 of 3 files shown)


vraravam (changes requested):
Please see the [updating guide](https://github.com/vraravam/recipes/blob/master/docs/updating.md) and follow the steps that are missing in this PR

vraravam (approved):
I haven't run these changes locally, but the changes seem fine

kytwb:
@all-contributors please add @ZwS for code

PR #776, opened 2021-11-28
Add RingCentral recipe

deadmeu:
This PR adds a recipe for the RingCentral unified communications app, which is the latest version of their teams messaging product (succeeding their classic messaging product _Glip_).

Both direct and indirect message badges have been implemented. Notifications don't work (I think the cause could be a permission issue) but that is something I'll look at in the future.

I've had a [basic recipe available](https://github.com/deadmeu/ferdi-ringcentral) for a while, but have only just bothered to clean things up to be published now.

This PR may resolve future issues like getferdi/ferdi#1388.


recipes/proton-mail/webview.js
```diff
@@ -2,10 +2,10 @@ module.exports = Ferdi => {
   const getMessages = () => {
     let unreadCount = 0;
     // Loop over all displayed counters and take the highest one (from the "All Mail" folder)
-    document.querySelectorAll('.navigation-counter-item').forEach(counterElement => {
+    for (const counterElement of document.querySelectorAll('.navigation-counter-item')) {
```
vraravam, line 5: Why are you editing the `proton-mail` recipe in this PR?

recipes/ringcentral/README.md
```diff
@@ -0,0 +1,2 @@
+# RingCentral for Ferdi
+This is a Ferdi recipe for the RingCentral unified communications app (formerly Glip).
```
vraravam, line 2: a `README.md` is no longer required - please remove

vraravam (changes requested):

recipes/proton-mail/webview.js
deadmeu: I think it was the pre-commit `eslint . --fix` command that made the change.

vraravam (approved):

kytwb:
@all-contributors please add @deadmeu for code

PR #780, opened 2021-12-01
Add recipe for Nextcloud News

csolisr:
Nextcloud News is a RSS news reading app for Nextcloud. This recipe is based on [`recipe-nextcloud-tasks`](https://github.com/meetfranz/recipe-nextcloud-tasks).


recipes/nextcloud-news/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "id": "nextcloud-news",
+  "name": "Nextcloud News",
+  "version": "1.1.1",
```
vraravam, line 4: For the first time a recipe is introduced, please start off with version `1.0.0`

recipes/nextcloud-news/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports = Ferdi => class NextcloudCarnet extends Ferdi {
+  buildUrl(url) {
```
vraravam, line 2: Rather than using this internal API, why not use the complete `serviceURL` in the `package.json`?

vraravam: I see that the `serviceURL` contains this suffix of `/apps/news` - why do we need this override of `buildURL`?

recipes/nextcloud-news/webview.js
```diff
@@ -0,0 +1,18 @@
+const _path = _interopRequireDefault(require('path'));
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
+module.exports = Ferdi => {
+  const getMessages = () => {
+    const selector = document.querySelector(
+      '.subscriptions-feed .app-navigation-entry-utils-counter',
+    );
+    const direct = selector.innerText ? parseInt(selector.innerText) : 0;
```
vraravam, line 10: Please use `Ferdi.safeParseInt` instead of `parseInt` - this already checks for `null` or `undefined` before parsing.

vraravam (changes requested):

recipes/nextcloud-news/package.json
csolisr: My bad, I was using another recipe as a template and forgot to fix something, BRB

recipes/nextcloud-news/index.js
csolisr: As I said, I was using another recipe as a template. Just to confirm, is this a proper replacement to the above?

```
module.exports = Ferdi => class NextcloudNews extends Ferdi {}
```

recipes/nextcloud-news/webview.js
csolisr: Thanks for the suggestion! I'm new with the Ferdi/Franz API system, so I didn't notice that `safeParseInt` existed already.

csolisr: Odd thing, whenever I call `Ferdi.safeParseInt` in that line, the debugger claims that the function is invalid. However, `Ferdi.setBadge` a few lines later is fetched properly. Checking further in, I find that in the `RecipeWebview.js` file that Ferdi pulls as a dependency, there is no such thing as a `safeParseInt` function! What am I doing wrong here?

vraravam: please see other recipes for examples. We have converted to `Ferdi.safeParseInt` all over the place

recipes/nextcloud-news/index.js
csolisr: Quick question: how do I ensure that the complete `serviceURL` is pulled in the function? I need to fetch a specific subpage, hence the `/apps/news/` suffix - when I remove that, Ferdi completely ignores the suffix

recipes/nextcloud-news/webview.js
csolisr: Understood, but I'm currently getting this message error whenever I attempt to use that function:

```
Uncaught TypeError: Ferdi.safeParseInt is not a function
    at RecipeWebview.getMessages [as loopFunc] (C:\Users\csoli\AppData\Roaming\Ferdi\recipes\dev\nextcloud-news\webview.js:10)
    at EventEmitter.<anonymous> (C:\Users\csoli\AppData\Local\Programs\ferdi\resources\app.asar\webview\lib\RecipeWebview.js:1)
    at EventEmitter.emit (events.js:210)
    at Object.onMessage (init.ts:50)
```

vraravam: Please see: https://github.com/vraravam/recipes/blob/master/docs/frontend_api.md#safeparseintstringtext

and this as an example: https://github.com/vraravam/recipes/blob/master/recipes/whatsapp/webview.js#L21

recipes/nextcloud-news/index.js
vraravam: not sure i understand. the `serviceUrl` is pulled completely always while rendering the page that the service recipe shows. If you comment out the whole function of `buildUrl` - and then run Ferdi, it should work.

csolisr: I need the site to direct to https://nextcloud.example.com/apps/news/ - but when I comment out the `buildUrl` function, the site directs to https://nextcloud.example.com/ instead (the dashboard page). For now my `package.json` file goes like this:

```
"config": {
    "hasNotificationSound": true,
    "hasIndirectMessages": false,
    "hasCustomUrl": true,
    "serviceURL": "https://{nextcloud}/apps/news/"
  }
```

Do I need any additional parameters and if so, why is the module fetching only the `{nextcloud}` parameter to form the URL instead of using the whole `serviceURL`?

csolisr: Never mind, I just found out that the `serviceURL` being used had to be updated in settings first to properly use the team ID.

recipes/nextcloud-news/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "id": "nextcloud-news",
+  "name": "Nextcloud News",
+  "version": "1.0.0",
+  "license": "MIT",
+  "repository": "https://github.com/csolisr/recipes/tree/master/recipes/nextcloud-news/",
+  "config": {
+    "hasNotificationSound": true,
+    "hasIndirectMessages": false,
```
vraravam, line 9: Since `hasIndirectMessages` defaults to `false`, please remove this line.

Look at [`docs/configuration.md`](https://github.com/vraravam/recipes/blob/master/docs/configuration.md)

vraravam (changes requested):

vraravam (approved):

kytwb:
@all-contributors please add @csolisr for code

keunes:
Thanks @csolisr for this work :) I just added it.

Since the header bar just displays the logo, the app icon, search, notifications and a personal menu that doesn't work (so you can't log out either), do you think we could simply hide the whole header?

The element following would need to be done:
* `<div id="content" class="app-news ng-scope" role="main" ng-app="News" ng-strict-di="" ng-controller="AppController as App">` gets `padding-top: 0;`
* `<header role="banner" id="header">` gets `display: none;`
* `<div id="app-navigation" ng-controller="NavigationController as Navigation" ng-hide="App.loading.isLoading('global')" class="ng-scope" style="">` gets `top: 0;`

Then you get a fullscreen news experience.

csolisr:
@keunes I was just using another Nextcloud app as a template, and I should probably do those fixes in a later iteration. I also have to reimplement the login for the rest of Nextcloud apps

keunes:
> I should probably do those fixes in a later iteration

I would also be happy to send a PR to the Ferdi repo.

> I also have to reimplement the login for the rest of Nextcloud apps

What login is implemented in the recipe? Doesn't it just rely on Nextcloud's login? (Sorry, maybe I misunderstood.)

csolisr:
> What login is implemented in the recipe? Doesn't it just rely on Nextcloud's login? (Sorry, maybe I misunderstood.)

Currently, several Nextcloud apps use a custom URL in the form of `https://{nextcloud}/apps/calendar/`, and unfortunately that's very prone to cause issues if the app is not located in a root path for example. So instead I'm using the teams setting to properly form the `{nextcloud}` path without the user accidentally modifying the template.

PR #783, opened 2021-12-03
Add Iris Messenger recipe

vbuzzano:
A recipe for Iris Messenger 
service url: https://iris.to
iris.to github: https://github.com/irislib/iris-messenger


recipes/iris-messenger/index.js
```diff
@@ -0,0 +1 @@
+module.exports = Ferdi => Ferdi;
```
vraravam, line 1: As of some commits done yesterday, this file (if empty or with only this "default" export) can be safely deleted.

recipes/iris-messenger/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": "iris-messenger",
+  "name": "Iris Messenger",
+  "version": "1.0.0",
+  "license": "MIT",
+  "config": {
+    "serviceURL": "https://iris.to",
+    "hasTeamId": false
```
vraravam, line 8: According to [the documentation](https://github.com/vraravam/recipes/blob/master/docs/configuration.md), `hasTeamId` is `false` by default. Please delete this line.

vraravam (changes requested):
Thanks for the contribution!

recipes/iris-messenger/package.json
vbuzzano: ah ok.  I thougth it was like Franz. I remember it was true by default.
I change asap :)

vbuzzano: ok i have done the change. don't remember if i need to do re-pull request or something else ?

recipes/iris-messenger/index.js
vraravam: Please delete this file.

vraravam (changes requested):

recipes/iris-messenger/package.json
vraravam: no need to recreate the PR. I am reviewing in this PR itself. Tx

recipes/iris-messenger/index.js
vbuzzano: I think you are wrong. Only empty webview.js files have been removed. And the package.js script still checks if the index.js file exists. Without it, we can't commit because an error is returned.

check: https://github.com/getferdi/recipes/blob/be4c5abcae4ef3d49bec8dba632381989af05311/scripts/package.js#L269-L274

vbuzzano: should i delete the file and remove package.js check ? or keep index.js in the recipe folder

vraravam: sorry - yes, my bad. I was thinking this was the `webview.js`

vraravam (approved):

kytwb:
@all-contributors please add @vbuzzano for code

PR #787, opened 2021-12-13
Indirect badge if title page starts with a star

### LINKED ISSUE #781: Nextcloud Talk: check for star in title indicating new messages in open chat
Using Nextcloud Talk in Ferdi works very well, but I'm missing one thing: when you're in a chat with someone and you minimize the window you don't get any indication that someone wrote you in the active chat. This is a Nextcloud thing because in this cases there is no Nextcloud notification generated. But in this cases the title of the website shows new messages with a star (*).

My idea would be to add something like this to the recipe of Nextcloud Talk (webview.js):

```
    if (document.title.startsWith("*")) {
      indirect++;
    }
```

This way you also get informed about new messages in the active chat.
I'm incrementing the indirect counter because this modification can't count the real number of new messages in the chat.
One drawback seems to be that you always get a indirect badge if you have disabled the Nextcloud Chat Notifications for the actual open chat. Therefore I wanted to ask the community if this modification wouldn't interfere to much with the logic of notifications.

cm-schl:
Resolves #781 

If the title of the Nextcloud Talk page starts with a star (*) increment the number of indirect messages.
Nextcloud shows the star indicator if you already are in an active chat and you get new notifications in this chat window.

A small drawback of this modification is the following:
If you get a direct notification from an other chat / by a mention in Nextcloud Talk and you open Ferdi to read the message SOMETIMES the direct badge gets removed and the indirect badge is shown for about a second.
This does not happen all the time because it depends of the refreshing rate of Ferdi and of the refreshing of Nextcloud itself.
In my personal opinion the fact that with this modification you get a indicator that in your active chat some message arrived seems to balance out this drawback.


vraravam (changes requested):
Please see the https://github.com/vraravam/recipes/blob/master/docs/updating.md documentation. You are missing the changes in the `package.json` and `all.json` files.

cm-schl:
Sorry, your completely right! I tried to update the version number - hope that this now is ok.

vraravam (approved):
Thanks - lgtm

cm-schl:
Hi @vraravam,

please excuse my question, but when / how the recipes on existing installations of Ferdi are updated when a pull request like this one was merged into master? I didn't find some comment in the documentation and I'm asking myself if I forgot to update some file. The Updater of Ferdi does not seem to find this new version of the recipe...

Thanks!

vraravam:
@cm-schl - I'm also not sure why this is happening.

The way the new recipes are referenced is as a **git-submodule**. During the process of the nightly builds, the reference to the submodule's git commit is [updated](https://github.com/getferdi/ferdi/blob/develop/.github/workflows/dependency-updates.yml#L47) and a new commit pushed. This new commit is then [picked up for packaging](https://github.com/getferdi/ferdi/blob/develop/.github/workflows/builds.yml#L47) inside the nightly Ferdi build.

I have downloaded version 5.6.4 and verified that the `/Applications/Ferdi.app/Contents/Resources/app.asar.unpacked/recipes/all.json` file contains the new version of `nextcloud-talk` which is `1.2.0`.

cm-schl:
@vraravam thanks for having a look to this and explaining the process.

I've tested on Windows the .msi (5.6.4), the .exe (5.6.4 and 5.6.5 nightly) and the tar.gz (5.6.5 nightly) and none of those contains version 1.2.0 of the nextcloud-talk recipe under `resources\app.asar.unpacked\recipes\`.

I saw also the following: entering directly to https://github.com/getferdi/recipes the version number in all.json is 1.2.0 for nextcloud-talk. When I enter to https://github.com/getferdi/ferdi the recipes are referenced to "recipes @ be4c5ab" which does not contain the modification (all.json = 1.1.1 for nextcloud-talk)

PR #788, opened 2021-12-16
Fix Rainloop unread counter

dpeukert:
jQuery isn't available on the RainLoop instance I'm using, so I've converted the webview.js script to vanilla JS.


kytwb:
Thanks for the contribution.

kytwb:
@all-contributors please add @dpeukert for code

PR #791, opened 2021-12-22
Use the Zoho Mail JS API instead of outdated HTML elements

bejonwe:
Hey all,

the current method for the count of unread e-mails is not working any more, as the  HTML of Zoho Mail changed. This is fixed by my commit, which now uses the JavaScript API of Zoho Mail instead of relying on CSS selectors to find a number in the HTML code.

Additionally, at the moment it is necessary to click a button on a splash screen to access the actual regional web interface of the Zoho account. This commit also adds an automatic redirect to the correct regional URL without requiring any interactions by the user.

Thanks for your work on this project and all the best,
Jonathan Weber

### DIFF
--- a/all.json
+++ b/all.json
@@ -2028,7 +2028,7 @@
     "featured": false,
     "id": "zoho",
     "name": "Zoho Mail",
-    "version": "1.1.0",
+    "version": "1.2.0",
     "icons": {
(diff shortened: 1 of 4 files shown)


kytwb:
@all-contributors please add @bejonwe for code

PR #812, opened 2022-01-02
Fix notifications, badge, and icon for The Lounge

briankendall:
This PR fixes a number of issues with The Lounge:

- Notifications weren't working
- The direct message badge was calculated based off of the number of channels with unread messages, rather than a tally of the unread messages among all the channels
- The badge didn't include unread messages in the current channel that arrived while the app didn't have focus
- The service icon was not using The Lounge's official logo

This PR closes https://github.com/getferdi/ferdi/issues/2372


briankendall:
This PR has a bug where notifications are generated for all new messages, rather than only highlighted ones. Will fix.

kytwb:
@briankendall thanks!

PR #817, opened 2022-01-07
Fix Infomaniak unread messages count

baukevdw:
When the last mail is read the indicator will stay at 1 since the count text is falsy when the unread count is 0.


kytwb:
@all-contributors please add @baukevdw for code

PR #824, opened 2022-01-20
Fix Google Drive serviceURL

nurse-the-code:
Fixed path to Google Drive. Old path led to a landing page with a link that would redirect to a web browser.


kytwb:
@nurse-the-code thanks

@all-contributors please add @nurse-the-code for code

PR #826, opened 2022-01-26
Fix Feedly counter script when no pending article to read

brunoccr:
Before fix when DOM not exists cause crash and not reset badge, now that's fixed.


kytwb:
@brunoccr thanks

@all-contributors please add @brunoccr for code

PR #846, opened 2022-03-16
Fixing zenhub recipe

alecive:
This is the first time I use the ZenHub recipe but it wasn't working out of the box. (Not sure if it was working in the past). As you can see from my commit, the link in the `serviceURL` field was broken, and my update (removing `/o/`, going from `workspace` to `workspaces` and from `boards` to `board`) fixed it.

Also, see screenshot below of a Zenhub instance with the correct link:
zenhub


kytwb:
@all-contributors please add @alecive for code

alecive:
Thank you @kytwb !

PR #859, opened 2022-04-06
Update Messenger notification counter

tamas646:
The Messenger's UI changed during the last months, and now we have the ability to count only non-muted chat channels.
Also I guess we don't need the old counting methods anymore...

If you wish to count muted channels too,
you can change the boolean variable back to true.

### DIFF
--- a/recipes/messenger/package.json
+++ b/recipes/messenger/package.json
@@ -1,7 +1,7 @@
 {
   "id": "messenger",
   "name": "Messenger",
-  "version": "1.3.4",
+  "version": "1.3.5",
   "license": "MIT",
   "config": {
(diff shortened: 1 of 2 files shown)


HugLifeTiZ:
Thank you for this! Could this be a starting point for also filtering out desktop notifications for muted chats? Right now, Messenger's web frontend pushes desktop notifications for muted chats.

tamas646:
Desktop notifications (meaning the 'text' notifications, not the sound) in the web frontend are not working at all for me (in Hungary). :cry:  There is no option for it in the settings page...
So unfortunately I can't answer your question.

weliveinferdi:
@all-contributors please add @tamas646 for code

PR #870, opened 2022-07-05
Fix The Lounge badge number so it ignores muted channels

briankendall:
This PR fixes a bug in the recipe for The Lounge where it would include muted channels when calculating its badge number.


briankendall:
My PR had a bug where The Lounge would still have a badge for networks that were muted, which I have now fixed.

weliveinferdi:
@briankendall thanks for the pull request!

